PR #15, opened 2019-03-11
Add alerts on free mem and free disk

mrunge:
- also add api status (if configured)
- alerts on high and low cpu usage


leifmadsen:
In progress

PR #25, opened 2019-06-20
Add deploy builder functionality

### LINKED ISSUE #24: Allow for SAF to be deployed into minishift
Allow for SAF to be deployed into different environment, namely minishift. We'll need to make certain manifests dynamic in order to support the different registry paths within OpenShift via Ansible and OpenShift via minishift.

leifmadsen:
In order to support multiple environments, we need to dynamically
create some of the manifests used in the deployment of Service
Assurance Framework.

The changes here allow for the spin up of SAF on both a standard
openshift-ansible installed system, and for a simple demo environment
on top of minishift.

We'll do this by providing some simple Ansible that generates a
handful of manifests using some defaults that can be overridden at the
console.

Closes #24


leifmadsen:
Just doing one more review locally before merging, but it seems to work for me other than one note about assignments to the nodes.

I might just remove the labeling to be honest since we're not really building things out to land on a specific node. I'll open that as a separate pull request though.

leifmadsen:
Needs #26 merged first and then this rebased against master.

leifmadsen:
**Current test plan**

```
minishift delete
minishift start
oc login -u system:admin
oc new-project sa-telemetry
oc create -f ~/6340056_cloudops-secret.yaml
openssl req -new -x509 -batch -nodes -days 11000 \
    -subj "/O=io.interconnectedcloud/CN=qdr-white.sa-telemetry.svc.cluster.local" \
    -out qdr-server-certs/tls.crt \
    -keyout qdr-server-certs/tls.key
oc create secret tls qdr-white-cert --cert=qdr-server-certs/tls.crt --key=qdr-server-certs/tls.key

ansible-playbook \
  -e "registry_path=$(oc registry info)" \
  -e "imagestream_namespace=$(oc project --short)" \
  deploy_builder.yml

# need to patch a node in order to allow the current version of the SGO to deploy a SG
oc patch node localhost -p '{"metadata":{"labels":{"application": "sa-telemetry", "node": "white"}}}'

./import-downstream.sh
./deploy.sh
watch -n5 oc get pods

./deploy.sh DELETE
watch -n10 oc get all
```

leifmadsen:
Ready for testing. I was able to successfully deploy and teardown a new environment with the current setup and test instructions.

CC: @mrunge @pkilambi

pkilambi:
👍  After I did the oc patch node all the pods are running as expected. Minor nit, I'm a bit skeptical for deploy.sh to wait and watch get pods. Might make sense to just let it finish and have user check the pod status later.

deploy/deploy.sh
```diff
@@ -85,14 +85,28 @@ delete() {
...
 if [ "$method" == "CREATE" ]; then
-    echo "  * [ii] Creating the operators" ; create ${operator_list[@]} && sleep 30
+    echo "  * [ii] Creating the operators" ; create ${operator_list[@]}
+    echo ""
+    echo "+-----------------------------------------------------------------------+"
+    echo "| Press Ctrl+C when prometheus-operator-1-<unique> is marked as Running |"
+    echo "+-----------------------------------------------------------------------+"
+    echo ""
+    trap ' ' INT
+    oc get pods -w
```
pkilambi: feel this could be a background task or perhaps we can add a verbose option and show only if thats passed. Not a blocker we can revisit later.

leifmadsen:
> +1 After I did the oc patch node all the pods are running as expected. Minor nit, I'm a bit skeptical for deploy.sh to wait and watch get pods. Might make sense to just let it finish and have user check the pod status later.

Yea it's a hack for now. It is required for now though because if you don't wait for the pods to finish deploying on the operator side, then the application side will fail (as we've seen). At that point, deleting isn't quite possible either because we have a similar problem with tearing down the operators too soon, and the applications don't get cleaned up properly.

PR #28, opened 2019-06-25
Drop nodeSelector from alertmanager roles

pkilambi:
Without this Alertmanager pod just sits in pending state due to FailedScheduling


leifmadsen:
This is already resolved in feature/minishift_deploy

pkilambi:
> This is already resolved in `feature/minishift`

I just pulled the latest and it was not fixed.

pkilambi:
see: https://github.com/redhat-service-assurance/telemetry-framework/blob/feature/minishift/deploy/roles/deploy.builder/templates/service-assurance/alertmanager/alertmanager.yaml#L7-L8

leifmadsen:
Oh I missed this was in the Ansible template. Will merge. Thanks!

PR #32, opened 2019-07-03
Set image stream imports to upstream

pleimer:
All images come from quay.io upstream versions

PR #33, opened 2019-07-03
Provide ability to install with upstream container images

pleimer:
Added quickstart script to automatically build and run upstream. This needs to be merged after #32

### DIFF
--- a/deploy/import-upstream.sh
+++ b/deploy/import-upstream.sh
@@ -1,15 +1,14 @@
+#!/bin/sh
 oc import-image prometheus:v2.11.104 --from=quay.io/prometheus/prometheus --confirm
 oc import-image prometheus-operator:latest --from=quay.io/coreos/prometheus-operator:v0.27.0 --confirm
 oc import-image prometheus-configmap-reloader:latest --from=quay.io/coreos/configmap-reload --confirm
-
 oc import-image prometheus-config-reloader:latest --from=quay.io/coreos/prometheus-config-reloader:v0.30.1 --confirm
 oc import-image prometheus-alertmanager:v0.15.0 --from=quay.io/prometheus/alertmanager --confirm
 
-oc import-image qdrouterd --from=quay.io/interconnectedcloud/qdrouterd --confirm
-oc import-image qdr-operator:latest --from=quay.io/interconnectedcloud/qdr-operator:1.0.0-beta2 --confirm
-
+oc import-image amq-interconnect:1.4-6 --from=quay.io/interconnectedcloud/qdrouterd:1.7.0 --confirm
+oc import-image amq-interconnect-operator --from=quay.io/interconnectedcloud/qdr-operator:1.0.0-beta2 --confirm
 
 oc import-image smart-gateway --from=quay.io/redhat-service-assurance/smart-gateway --confirm
 oc import-image smart-gateway-operator --from=quay.io/redhat-service-assurance/smart-gateway-operator --confirm
 
-oc set image-lookup prometheus prometheus-operator prometheus-configmap-reloader prometheus-config-reloader prometheus-alertmanager qdrouterd qdr-operator smart-gateway smart-gateway-operator
+oc set image-lookup prometheus prometheus-operator prometheus-configmap-reloader prometheus-config-reloader prometheus-alertmanager amq-interconnect amq-interconnect-operator smart-gateway smart-gateway-operator
--- a/deploy/import-upstream.sh
+++ b/deploy/import-upstream.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-oc import-image prometheus:v2.11 --from=quay.io/openshift/origin-prometheus:v3.11 --confirm
+oc import-image prometheus:v2.11.0 --from=quay.io/openshift/origin-prometheus:v3.11 --confirm
 oc import-image prometheus-operator:latest --from=quay.io/openshift/origin-prometheus-operator:v3.11 --confirm
 oc import-image prometheus-configmap-reloader:latest --from=quay.io/openshift/origin-configmap-reload:v3.11 --confirm
 oc import-image prometheus-config-reloader:latest --from=quay.io/openshift/origin-prometheus-config-reloader:v3.11 --confirm
--- a/deploy/import-upstream.sh
+++ b/deploy/import-upstream.sh
@@ -1,14 +1,14 @@
 #!/bin/sh
-oc import-image prometheus:v2.11.104 --from=quay.io/prometheus/prometheus --confirm
-oc import-image prometheus-operator:latest --from=quay.io/coreos/prometheus-operator:v0.27.0 --confirm
-oc import-image prometheus-configmap-reloader:latest --from=quay.io/coreos/configmap-reload --confirm
-oc import-image prometheus-config-reloader:latest --from=quay.io/coreos/prometheus-config-reloader:v0.30.1 --confirm
-oc import-image prometheus-alertmanager:v0.15.0 --from=quay.io/prometheus/alertmanager --confirm
+oc import-image prometheus:v2.11 --from=quay.io/openshift/origin-prometheus:v3.11 --confirm
+oc import-image prometheus-operator:latest --from=quay.io/openshift/origin-prometheus-operator:v3.11 --confirm
+oc import-image prometheus-configmap-reloader:latest --from=quay.io/openshift/origin-configmap-reload:v3.11 --confirm
+oc import-image prometheus-config-reloader:latest --from=quay.io/openshift/origin-prometheus-config-reloader:v3.11 --confirm
+oc import-image prometheus-alertmanager:v0.15.0 --from=quay.io/openshift/origin-prometheus-alertmanager:v3.11 --confirm
 
 oc import-image amq-interconnect:1.4-6 --from=quay.io/interconnectedcloud/qdrouterd:1.7.0 --confirm
-oc import-image amq-interconnect-operator --from=quay.io/interconnectedcloud/qdr-operator:1.0.0-beta2 --confirm
+oc import-image amq-interconnect-operator:latest --from=quay.io/interconnectedcloud/qdr-operator:1.0.0-beta2 --confirm
 
-oc import-image smart-gateway --from=quay.io/redhat-service-assurance/smart-gateway --confirm
-oc import-image smart-gateway-operator --from=quay.io/redhat-service-assurance/smart-gateway-operator --confirm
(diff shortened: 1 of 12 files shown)


deploy/roles/deploy.builder/templates/operators/qdrouterd/operator.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       containers:
         - name: qdr-operator
           # Replace this with the built image name
-          image: {{ registry_path }}/{{ imagestream_namespace }}/amq-interconnect-operator
+          image: {{ registry_path }}/{{ imagestream_namespace }}/qdr-operator
```
csibbitt: :-1: AFAICT this will no longer work with the downstream image streams so we can't merge this. Since this is referring to our local imagestream name, I think we can continue to call it amq-interconnect-operator on our side.

leifmadsen: Agreed. The image stream names should probably just remain exactly the same as the downstream version. There is probably no reason to change anything other than the `--from=...` part

deploy/import-upstream.sh
```diff
@@ -1,2 +1,14 @@
-oc import-image coreos/prometheus-operator:v0.25.0 --from=quay.io/coreos/prometheus-operator:v0.25.0 --confirm
-oc import-image coreos/configmap-reload:v0.0.1 --from=quay.io/coreos/configmap-reload:v0.0.1 --confirm
+#!/bin/sh
+oc import-image prometheus:v2.11.104 --from=quay.io/prometheus/prometheus --confirm
+oc import-image prometheus-operator:latest --from=quay.io/coreos/prometheus-operator:v0.27.0 --confirm
+oc import-image prometheus-configmap-reloader:latest --from=quay.io/coreos/configmap-reload --confirm
+oc import-image prometheus-config-reloader:latest --from=quay.io/coreos/prometheus-config-reloader:v0.30.1 --confirm
+oc import-image prometheus-alertmanager:v0.15.0 --from=quay.io/prometheus/alertmanager --confirm
+
+oc import-image amq-interconnect:1.4-6 --from=quay.io/interconnectedcloud/qdrouterd:1.7.0 --confirm
```
csibbitt:
```
<leifmadsen> yea... what we really want is 1.7.0...
<leifmadsen> that's what we have downstream
<leifmadsen> so yea... the version scheme frankly sucks
<leifmadsen> 1.3 == 1.5, 1.4 == 1.7
<leifmadsen> amq-interconnect vs qdrouterd
<csibbitt> Oh yeah? haha. Okay! That I can handle. If the downstream amq-interconnect:1.4  is equivalent to upstream qdrouterd:1.7.0  (I'll just take your word for it unless you have an easy reference I can read?) - then we're in good shape.
<leifmadsen> I just kind of know from memory
```

deploy/import-upstream.sh
```diff
@@ -1,2 +1,14 @@
-oc import-image coreos/prometheus-operator:v0.25.0 --from=quay.io/coreos/prometheus-operator:v0.25.0 --confirm
-oc import-image coreos/configmap-reload:v0.0.1 --from=quay.io/coreos/configmap-reload:v0.0.1 --confirm
+#!/bin/sh
+oc import-image prometheus:v2.11.104 --from=quay.io/prometheus/prometheus --confirm
```
leifmadsen: Let's pick a specific tag rather than relying on `latest`. Comment goes for all the components. We'll need to do the same thing for the downstream script as well.

csibbitt:
# Testing
```
[csibbitt@laptop deploy (qsupstream) ]$ oc delete project sa-telemetry
[csibbitt@laptop deploy (qsupstream) ]$ ./quicktest_upstream.sh 
[...]
NAME                                      READY     STATUS    RESTARTS   AGE
alertmanager-sa-0                         2/2       Running   0          1m 
prometheus-operator-1-kk4gn               1/1       Running   0          1m 
prometheus-white-0                        3/3       Running   1          1m 
qdr-operator-7c785c455d-pbk6h             1/1       Running   0          1m 
qdr-white-6fb67d46d6-czqnr                1/1       Running   0          1m 
smart-gateway-operator-5c96899578-dxptk   1/1       Running   0          1m 
white-smartgateway-1-bn5jd                1/1       Running   0          1m 
```

No other functional testing (I'm working on figuring out a lightweight way to smoke test SAF, hopefully coming "soon")

deploy/import-upstream.sh
```diff
@@ -1,2 +1,14 @@
...
+oc import-image prometheus-operator:latest --from=quay.io/coreos/prometheus-operator:v0.27.0 --confirm
```
leifmadsen: Our downstream version is 0.25.0. I assume there is nothing wrong with 0.27.0 but I wonder if we want to sync across for a semi-matching experience between downstream and upstream?

deploy/quickstart_upstream.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# import downstream container images
+./import-upstream.sh
+
+# deploy the environment (requires interaction)
+./deploy.sh
+watch -n5 oc get pods
+
+# teardown the environment when done (requires interaction)
+./deploy.sh DELETE
```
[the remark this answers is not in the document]
leifmadsen: I say we just delete that for now. Could just replace it with an `echo "To teardown the environment, run:  ./deploy.sh DELETE"`

deploy/import-upstream.sh
pleimer: Should we tag everything uniformly, including latest and specific versions?

deploy/import-upstream.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/sh
-oc import-image prometheus:v2.11.104 --from=quay.io/prometheus/prometheus --confirm
-oc import-image prometheus-operator:latest --from=quay.io/coreos/prometheus-operator:v0.27.0 --confirm
-oc import-image prometheus-configmap-reloader:latest --from=quay.io/coreos/configmap-reload --confirm
-oc import-image prometheus-config-reloader:latest --from=quay.io/coreos/prometheus-config-reloader:v0.30.1 --confirm
-oc import-image prometheus-alertmanager:v0.15.0 --from=quay.io/prometheus/alertmanager --confirm
+oc import-image prometheus:v2.11 --from=quay.io/openshift/origin-prometheus:v3.11 --confirm
```
leifmadsen: @csibbitt @pleimer I've migrated to using the OpenShift organization for the container images, and all are now tagged with the `v3.11` image tag.

deploy/roles/deploy.builder/templates/operators/qdrouterd/operator.yaml
leifmadsen: Fixed and tested.

deploy/import-upstream.sh
leifmadsen: At this point I think I've resolved this.

deploy/import-upstream.sh
leifmadsen: Resolved.

leifmadsen:
OK everything has been updated to something resembling sanity in the `import-upstream.sh` script. I've also ran the SmokeTest which worked successfully with these new images!

deploy/import-upstream.sh
csibbitt: Follow up on this based on version discussions we're having. If we're okay with trying to use "latest" for internal tags for now because the source tag is more specific anyways, should we use `:latest` here, too? IIUC, the prometheus operator demands :vX.Y.Z versions for it's containers, but it seems like we could use 'latest' for everything else. 

I'm still not 100% comfortable with "latest" because it implies a certain behavior that you might not actually get (for example if you override the imagePullPolicy to something other than Always, or whether you do or do not have an Image Change Trigger in your deployment). I think this is part of why some people use a "master" tag. I'm okay with keeping it for now because I'm not familiar enough with ImageStreams to make a solid recommendation today, but if we can then we should at least try to be 100% consistent.

deploy/import-upstream.sh
```diff
@@ -1,2 +1,14 @@
...
+oc import-image prometheus:v2.11.0 --from=quay.io/openshift/origin-prometheus:v3.11 --confirm
+oc import-image prometheus-operator:latest --from=quay.io/openshift/origin-prometheus-operator:v3.11 --confirm
+oc import-image prometheus-configmap-reloader:latest --from=quay.io/openshift/origin-configmap-reload:v3.11 --confirm
+oc import-image prometheus-config-reloader:latest --from=quay.io/openshift/origin-prometheus-config-reloader:v3.11 --confirm
+oc import-image prometheus-alertmanager:v0.15.0 --from=quay.io/openshift/origin-prometheus-alertmanager:v3.11 --confirm
+
+oc import-image amq-interconnect:1.4-6 --from=quay.io/interconnectedcloud/qdrouterd:1.7.0 --confirm
+oc import-image amq-interconnect-operator:latest --from=quay.io/interconnectedcloud/qdr-operator:1.0.0-beta2 --confirm
+
+oc import-image smart-gateway --from=quay.io/redhat-service-assurance/smart-gateway:latest --confirm
```
csibbitt: These two might as well get `:latest` on the left hand side for consistency.

Not sure we want to use `:latest` on the right hand side, but still okay with approving it for now.

deploy/import-upstream.sh
csibbitt: Follow up on this based on version discussions we're having. If we're okay with trying to use "latest" for internal tags for now because the source tag is more specific anyways, should we use :latest here, too? IIUC, the prometheus operator demands :vX.Y.Z versions for it's containers, but it seems like we could use 'latest' for everything else.

I'm still not 100% comfortable with "latest" because it implies a certain behavior that you might not actually get (for example if you override the imagePullPolicy to something other than Always, or whether you do or do not have an Image Change Trigger in your deployment). I think this is part of why some people use a "master" tag. I'm okay with keeping it for now because I'm not familiar enough with ImageStreams to make a solid recommendation today, but if we can then we should at least try to be 100% consistent.

deploy/import-upstream.sh
leifmadsen: Done

deploy/import-upstream.sh
leifmadsen: Notes have been added to help clarify why we're doing things the way we're doing it. Additional work is likely required here, but I'm hoping that our eventual migration to OLM will result in a lot of this stuff just taking care of itself.

leifmadsen:
[Implement `import-upstream.sh` script for telemetry-framework](https://trello.com/c/QlXOuZ8f/588-implement-import-upstreamsh-script-for-telemetry-framework)

PR #36, opened 2019-07-10
Minor deploy.sh improvements

csibbitt:
* Removed one of the interactive steps
  * A non-interactive setup is key for CI usage
* Removed some un-needed loops
* Now passes shellcheck

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -62,43 +62,33 @@ declare -a application_list=(
 )
 
 create() {
-    # create our objects by building a list
-    built_list=""
     object_list=("$@")
-    for spec in ${object_list[@]}; do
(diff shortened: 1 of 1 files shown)


deploy/deploy.sh
```diff
@@ -62,51 +62,41 @@ declare -a application_list=(
...
 create() {
-    # create our objects by building a list
-    built_list=""
     object_list=("$@")
-    for spec in ${object_list[@]}; do
-        built_list+=" -f $spec"
-    done
-
-    oc create --save-config=true $built_list
+    # shellcheck disable=SC2068
```
csibbitt: The existing code already failed SC2068 - the only alternative here is to put the `-f` on every line of our array, which seems silly.

PR #39, opened 2019-07-17
Fixed the deployment to pass tests and did some label cleanup 

csibbitt:
* Removed redundant ServiceMonitor
* Fixed 'alertmanager' route
  * No backends matched the label and it seemed redundant
  * Removed it and pointed the route to `alertmanager-operated` instead
* Documented a proposed label set that we should stick to
* Eliminated several non-conformant labels
* Fixed some markdown lint
* Moved some waits from travis.yml to deploy.sh because humans need to wait too

* I've found two places where the labels intersect with the operator

  * Node labels
    * https://github.com/redhat-service-assurance/smart-gateway-operator/blob/7386c50807c09fb2229a35c9063c464213539078/roles/smartgateway/tasks/main.yml#L52
    * This needs a node labelled as `application: sa-telemetry`, `node: white`
    * Which is what we do here: https://github.com/redhat-service-assurance/telemetry-framework/blob/3562d4492a22fdd784c7689fd06cc50835d0a779/deploy/quicktest_upstream.sh#L17
    * I propose that this change to `app: smartgateway`, `sa-affinity: white`
    * Do we even need this? Why is the smartgateway dependent on a specifically
      labelled node when other components are not?
    * If necesarry, we can handle this in a new PR because of the dependency

  * ServiceMonitor labels
    * ServiceMonitor `smartgateway: white` label: https://github.com/redhat-service-assurance/smart-gateway-operator/blob/7386c50807c09fb2229a35c9063c464213539078/roles/smartgateway/tasks/main.yml#L84
    * Prometheus `smartgateway: white` selector: https://github.com/redhat-service-assurance/telemetry-framework/blob/3562d4492a22fdd784c7689fd06cc50835d0a779/deploy/service-assurance/prometheus/prometheus.yaml#L35
    * I propose these change to `app: smartgateway`

* The other labels used in the operator appear to be self-contained
  * Should be able to fix them there without touching this repo
  * Pod `app: prometheus-white` label: https://github.com/redhat-service-assurance/smart-gateway-operator/blob/7386c50807c09fb2229a35c9063c464213539078/roles/smartgateway/tasks/main.yml#L25
  * Service `smartgateway: white` label: https://github.com/redhat-service-assurance/smart-gateway-operator/blob/7386c50807c09fb2229a35c9063c464213539078/roles/smartgateway/tasks/main.yml#L65
  * ServiceMonitor `smartgateway: white` selector: https://github.com/redhat-service-assurance/smart-gateway-operator/blob/7386c50807c09fb2229a35c9063c464213539078/roles/smartgateway/tasks/main.yml#L88
  * Note that the two above are actually not related to the ServiceMonitor
    label & Prometheus selector mentioned in the previous section
  * I propose making all of these `app: smartgateway, sa-affinity:white`


csibbitt:
[Get SAF smoke test working in Travis CI](https://trello.com/c/LzhKvEUN/595-get-saf-smoke-test-working-in-travis-ci)

deploy/README.md
```diff
@@ -1,4 +1,4 @@
-# Deployment using Operators
+# SAF Deployment using Operators
```
leifmadsen: Should probably just call this out in the first instance:  Service Assurance Framework

leifmadsen: Of course I'm not blocking this for this change :)

deploy/deploy.sh
```diff
@@ -50,17 +49,24 @@ declare -a application_list=(
...
     'service-assurance/alertmanager/alertmanager.yaml'
-    'service-assurance/alertmanager/service.yaml'
     'service-assurance/alertmanager/route.yaml'
 )
 
+declare -a crds_to_wait_for=(
+    'alertmanagers.monitoring.coreos.com'
+    'prometheuses.monitoring.coreos.com'
+    'prometheusrules.monitoring.coreos.com'
+    'qdrs.interconnectedcloud.github.io'
```
leifmadsen: In the future this will change to `interconnects` I believe. No change for now, just noting it.

deploy/service-assurance/prometheus/prometheus.yaml
```diff
@@ -18,15 +16,15 @@ spec:
       port: metrics
   podMetadata:
     labels:
-      sa-app: prometheus-white
+      app: prometheus
+      sa-affinity: white
```
leifmadsen: Our discussion proposes that we drop affinity

deploy/README.md
csibbitt: Fixed

deploy/service-assurance/prometheus/prometheus.yaml
csibbitt: fixed

PR #40, opened 2019-07-18
Initial attempt at travis smoke tests

leifmadsen:


leifmadsen:
I was looking at the difference between push and pr, but I guess I misunderstood what was passing and what was failing. I have reverted the merge into master, and opened another PR out of the original branch that had been previously merged. Guess we need to get both CI tests to pass.

leifmadsen:
https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests

leifmadsen:
OK so I just merged this to master where it failed, reverted the change to master, created a new PR based on the branch that I had just merged, and then both tests passed on the re-test.

Going to run both tests again (`pr` and `push`) and then see what happens. If that passes, then I suppose it should be merged to master?

Only other reason would be that this branch needs to be rebased.

PR #43, opened 2019-07-23
Change label `smartgateway: white` to `app: smart-gateway`

csibbitt:
This is compatible with the current smart-gateway-operator master branch: https://github.com/redhat-service-assurance/smart-gateway-operator/blob/master/roles/smartgateway/tasks/main.yml#L81


csibbitt:
[[SAF] Resolve issues with smoketest not completing when running downstream vs upstream image imports](https://trello.com/c/XeWVNVTW/594-saf-resolve-issues-with-smoketest-not-completing-when-running-downstream-vs-upstream-image-imports)

PR #49, opened 2019-08-15
This adds the automated performance test to SAF for use with minishift...

pleimer:
The performance test deploys two pods that execute test logic and host Grafana for data analysis. Collectd-tg is used to pump simulated network data through the smart gateway and into the prometheus database at user specified rates, intervals, and sizes. Grafana dashboards are automatically generated to graphically display queries to the database during the time of each test. 

In addition, files within the tests/ directory have been divided into separate folders to minimize confusion.


tests/performance-test/Dockerfile
```diff
@@ -0,0 +1,24 @@
+#--- Build SAF performance test ---
+FROM golang:1.12.7
+WORKDIR /go/src/performance-test/
+
+COPY ./dashboard.go .
```
leifmadsen: I think this can be combined to a single line. Note a big deal.

`COPY ./dashboard.go ./main.go ./parser.go .`

tests/performance-test/README.md
```diff
@@ -0,0 +1,58 @@
+# SAF Performance Test
+
+## Introduction
+The performance test provides an automated environment in which to to run stress tests on the SAF locally using Minishift. Collectd-tg is used to simulate extensive netwrok traffic to pump through SAF. Because Minishift only supports a single node at a time, this test demonstrates the limits of SAF in a constrained environment. Test scenarios are manually configured in a yaml file and results can be analyzed in a series of grafana dashboards.
```
leifmadsen: Wrap to 80 chars wide please

tests/performance-test/dashboard.go
```diff
@@ -0,0 +1,121 @@
...
+
+type Dashboard struct {
+	client        *sdk.Client
+	board         *sdk.Board
+	grafUrl       string
+	promUrl       string
+	panelTemplate []byte
+}
+
+func findDsWithUrl(matchURL string, dsList []sdk.Datasource) (*sdk.Datasource, error) {
```
leifmadsen: Could use some golang documentation. See https://blog.golang.org/godoc-documenting-go-code

pleimer:
It looks like the travis ci is failing because I moved the location of the smoketest.

tests/performance-test/deploy/config/test-configs.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: SAF Performance Test 1
+  spec:
+    value-lists: 10000
+    hosts: 5000
+    plugins: 100
+    interval: 1
+    length: 20
+    queries:
+      - rate(sa_collectd_total_amqp_processed_message_count[10s])
+      - sa_collectd_cpu_total
```
leifmadsen: Can you change this to another metric? The `sa_collectd_cpu_total` I'm pretty sure only shows you the total number of available cores, which probably isn't what you want here. I'd suggest maybe listing a 2 or 3 extra metrics. You have the number of processed message counts. It would be useful to also see the actual CPU load, the memory load, and maybe the disk IO or something like that.

If you have a running environment (I just tore all mine down) we can look at the list and pick 2-3 of them that make sense here.

tests/performance-test/deploy/performance-test-job.yml
```diff
@@ -0,0 +1,52 @@
...
+spec:
+  template:
+    metadata:
+      label:
+        name: saf-performance-test
+    spec:
+      restartPolicy: Never
+      containers:
+        - name: performance-test
+          image: 172.30.1.1:5000/sa-telemetry/performance-test:dev
```
leifmadsen: Enhancement here (future) would be to take these files and make them Ansible templates so that it could generate the registry dynamically like we do prior to the use of `deploy.sh`. Not necessary now, but something that would move this out of the minishift environment.

Don't do that now obviously. Future work.

tests/performance-test/deploy/performance-test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
leifmadsen: Validate this passes shellcheck.

tests/performance-test/deploy/scripts/performance-test-entrypoint.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
leifmadsen: Validate this passes shellcheck.

tests/performance-test/deploy/scripts/unit-test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
leifmadsen: Validate this passes shellcheck.

tests/performance-test/main.go
```diff
@@ -0,0 +1,133 @@
+package main
```
leifmadsen: Same comment before about godocs.

leifmadsen:
@paramite can you review the golang code?

tests/performance-test/deploy/config/test-configs.yml
pleimer: Yeah I can add some better defaults - I really just have them there as an example. What do you think would be more useful default queries?

tests/performance-test/dashboard.go
```diff
@@ -97,7 +103,7 @@ func (d *Dashboard) AddGraph(title string, query string) error {
 	return err
 }
 
-//Update dashboard if it alread exists in Grafana. Otherwise, create a new one.
+//Update updates the dashboard if it alread exists in Grafana. Otherwise, create a new one.
```
leifmadsen: s/alread/already

pleimer:
> Haven't had a chance to test all this, but it seems legit :)

Haha I think it is! I want to try and get one additional small thing working before I squash and rebase

PR #54, opened 2019-08-29
Basic multicloud support

csibbitt:
* Changed "white" to "cloud1" where appropriate for clarity
* QDR configured for entire collectd/* prefix as multicast
  * Included change in infrared deployment template as well
  * Eventually this will be default, see https://review.opendev.org/#/c/678961/
* Templated smoketest job and collectd
* Smoketest now handles its own smart gateways
  * Instantiates one smart gateway and one collectd per "cloud"
* Each test pod checks for only it's own messages
  * Also ensures they come from the right smart gateway


csibbitt:
Documentation is nearly complete, will follow up with it soon.

# Testing
I ran this with up to 4 simulated clouds (`./smoketest.sh 4`) and it worked each time. I confirmed from the test pod logs and by manually exploring in prometheus that each pod successfully delivered messages though the correct smart gateway.

During testing I did see test pods fail (though with the retries they still succeeded within the timeout and passed the test). I investigated the failed pods and found that they delivered messages to the smart gateway and it processed them, but sometimes the prometheus ServiceMonitor is slow to add new targets, so they are not always scraped soon enough to pass the test. This was the cause of the retries, and once the target gets added to Prometheus, one of the pods would pass. I briefly explored trying to wait on the targets to appear but got a permission error from the prometheus API trying to query targets and did not sink any more time into it.

csibbitt:
[Test multiple IR-based OpenStack clouds connecting to a single QuickLab instance](https://trello.com/c/j9W5N853/629-test-multiple-ir-based-openstack-clouds-connecting-to-a-single-quicklab-instance)

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
+# Configuring Service Assurance Framework for Multiple Clouds
+
+Multiple Openstack clouds can be configured to target a single instance of SAF.
+There are a few steps to get this set up:
+
+1. Plan the AMQP address prefixes to use for each cloud
+1. Deploy metrics and notification SmartGateways for each cloud to listen on the
```
leifmadsen: > Deploy metrics and events consumer Smart Gateways

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
...
+  corresponding address prefixes
+1. Configure each cloud to send it's metrics and notifications to SAF on the
```
leifmadsen: s/notifications/events

Only suggesting this change since these are the terms we've been using.

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
+# Configuring Service Assurance Framework for Multiple Clouds
+
+Multiple Openstack clouds can be configured to target a single instance of SAF.
```
leifmadsen: s/SAF/Service Assurance Framework (SAF)/

Usually when using an acronym it's called out in full the first time, then defined and used for the rest of the document.

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
...
+  corresponding address prefixes
+1. Configure each cloud to send it's metrics and notifications to SAF on the
+  correct address
+
+## AMQP addresses
+
+By default, OSP nodes are configured to send data to the `collectd/telemetry`
+and `collectd/notify` addresses on the AMQP bus; and SAF is configured to
+listen on those addresses for monitoring data. In order to support multiple
+clouds and have the ability to easily identify which cloud genereated which
```
leifmadsen: generated

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
...
+
+* collectd/cloud1-telemetry
+* collectd/cloud1-notify
+* collectd/cloud2-telemetry
+* collectd/cloud2-notify
+* collectd/us-east-1-telemetry
+* collectd/us-west-3-telemetry
+* ...etc
+
+## Deploying SmartGateways
```
leifmadsen: We've been using "Smart Gateway" vs "SmartGateway" in prior documentation.

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
...
+
+Two SmartGateways (one for metrics, one for events) need to be deployed
```
leifmadsen: Technically you could just get metrics or events, although I'm not sure that pedantry is necessary here :)

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
...
+for each cloud, configured to listen on the correct AMQP address. For example:
+
+```yaml
+apiVersion: smartgateway.infra.watch/v1alpha1
+kind: SmartGateway
+metadata:
+  name: cloud1-telemetry
+spec:
+  amqp_url: qdr-white.sa-telemetry.svc.cluster.local:5672/collectd/cloud1-telemetry
+  serviceType: metrics
```
leifmadsen: So the Ansible Operator will convert the camel case to snake case for you, but I find it weird that our examples here are a mix of snake and camel case.

In the manifests I've just been generating them into snake case so there is no reliance on a translation mechanism.

https://github.com/operator-framework/operator-sdk/blob/master/doc/ansible/user-guide.md#define-the-memcached-spec

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
...
+spec:
+  amqp_url: qdr-white.sa-telemetry.svc.cluster.local:5672/collectd/cloud2-notify
+  serviceType: events
+```
+
+## Openstack configuration
+
+In order to label traffic according to it's cloud of origin, the collectd
+configuration has to be updated to have cloud-specific instance names. This is
+usually accomplished by editting your triple-o configuration to have the
```
leifmadsen: * editing
* TripleO

deploy/README-multicloud.md
```diff
@@ -0,0 +1,100 @@
...
+following CollectdAmqpInstances.
```
leifmadsen: wrap CollectedAmqpInstances in backticks

deploy/README.md
```diff
@@ -21,40 +21,12 @@ currently a work in a progress.
...
-    ./import-downstream.sh
-
-    # deploy the environment (requires interaction)
-    ./deploy.sh
-    watch -n5 oc get pods
-
-    # teardown the environment when done (requires interaction)
-    ./deploy.sh DELETE
-    watch -n10 oc get all
+This script can be found here: https://github.com/redhat-service-assurance/telemetry-framework/blob/master/deploy/quickstart.sh
```
leifmadsen: The script

deploy/README-multicloud.md
csibbitt: Fixing this, but I must admit that I got a bit confused looking into which is actually used because the CRD shows camelCase: https://github.com/redhat-service-assurance/telemetry-framework/blob/b15c678fc60b3545d95b85739d3a2af5109b35a1/deploy/operators/smartgateway/crds/smartgateway_v1alpha1_smartgateway_crd.yaml#L37  whereas the CRs we actually deploy has snake_case: https://github.com/redhat-service-assurance/telemetry-framework/blob/b15c678fc60b3545d95b85739d3a2af5109b35a1/deploy/operators/smartgateway/crds/smartgateway_v1alpha1_smartgateway_cr.yaml#L7 

To make things more confusing, the snake_case keys from the CR get translated BACK into camelCase for the config: https://github.com/redhat-service-assurance/smart-gateway-operator/blob/1ea1f822d331a96fc62f11045ed2795255884aac/roles/smartgateway/templates/events-configmap.yaml.j2#L13  I think Ansible is the odd one out here, and it's going to be confusing no matter what we do - your recommendation to avoid the auto-translation seems reasonable, but I'm not sure we've been consistent.

As an aside, I'm also noticing that the CRD still also says "AMQPMetricUrl" - so is it just not used when validating a resource going into the cluster?

deploy/operators/smartgateway/crds/smartgateway_v1alpha1_smartgateway_cr.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: smartgateway.infra.watch/v1alpha1
 kind: SmartGateway
 metadata:
-  name: white
+  name: cloud1
 spec:
   size: 1
   amqpmetric_url: qdr-white-normal.sa-telemetry.svc.cluster.local:5672/collectd/telemetry
```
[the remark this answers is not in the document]
leifmadsen: Yea those files are generated as part of the operator-sdk framework. I think we're going to find that they are either 1) not necessary or 2) are necessary as part of OLM :)

deploy/README-multicloud.md
leifmadsen: Honestly I want to figure out how to have the CRD generated for us based on the actual code implementation. At this point the CRD is actually quite out of date based on the most recent set of changes. I'm not sure how to handle this at this point. The alternative is to just strip out all the parameters listed because I'm not even sure the validation happens since you'd expect to have started to see errors deploying once we changed from `AMQPMetricsUrl` to `AMQPUrl`.

leifmadsen: And yea, I agree that the Ansible operator approach is pretty confusing...

I don't have a strong feeling in either case, because I see validity to both approaches: 1) just set the inbound values as what you would expect them to be after inbound and outbound translations, or 2) set the values to what Ansible is going to expect to convert them to anyways so that we don't cause some sort of break, but results in confusion when the outbound translation converts it back to camelCase.

deploy/README.md
```diff
@@ -129,6 +101,11 @@ Then simply run the `deploy.sh` script. You will need to follow the
 instructions during the script as it will pause waiting for the successful
 completion at a couple of steps.
 
+## Multi-cloud support
+
+It is possible to deploy one intance of SAF and attach multiple openstack clouds
+to it. See the [Multicloud README](README-multicloud.md) for details.
```
leifmadsen: I guess pick _Multi-cloud_ or _Multicloud_

leifmadsen: And I agree that I'm being far too picky for what is likely temporary documentation :)

PR #62, opened 2019-09-09
Merge deploy/README.md to main docs

csibbitt:
* Most of it is covered more succinctly in main docs
* Added quickstart and container source sections
  * Both only set to render for the "upstream" target so they don't appear in the official downstream docs (we don't want to support quickstart.sh or upstream containers for RH customers)
* Moved the labels section to it's own file for now. Eventually we'll have reference documentation of some sort and it can go there.


csibbitt:
[SAF Docs refactor](https://trello.com/c/SSUAZqyc/655-saf-docs-refactor)

PR #63, opened 2019-09-17
[WIP] Add smoke tests for events

leifmadsen:
Working towards adding smoke tests that will check for an index being created in ElasticSearch.


leifmadsen:
[Create test for events in smoketest](https://trello.com/c/yE75uVPq/668-create-test-for-events-in-smoketest)

PR #64, opened 2019-09-19
Update TripleO configuration instructions to include details on DCN

elfiesmelfie:
Add note on how to configure tripleo for distributed compute nodes.

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -345,9 +345,10 @@ The SSL certificates for the `MetricsQdr` service is configured to generate only
 
 [id=`configuration-for-distributed-compute-nodes_{context}`]
 === Configuration for distributed compute nodes
-If you are deploying compute nodes on networks different from the controllers, some additional changes are required to allow data transported via AMQ INterconnect to reach the SAF server instance.
+If you are deploying compute nodes on networks different from the controllers, such as in the case of distributed compute nodes (DCN),
+some additional changes are required to allow data transported via AMQ Interconnect to reach the SAF server instance.
 
-The following configuration exambple needs to be added under the `ExtraConfig` section for each compute node.
+The following configuration example needs to be added under the `ExtraConfig` section for each compute node.
 In the example, `internal_apiN` is the network that the compute node N is connected to, where N represents a value different from the default internal API network name.
 
 [source,yaml]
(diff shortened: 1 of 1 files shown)


docs/README.adoc
```diff
@@ -343,6 +343,20 @@ openstack overcloud deploy \
 [NOTE:] 
 The SSL certificates for the `MetricsQdr` service is configured to generate only for the `InternalApi` network but the default Ceph role does not use the `InternalApi` network. To deploy SAF client when InternalTLS is enabled, use this workaround: pass the custom Ceph role, which has `InternalApi` network, to `openstack overcloud deploy` when InternalTLS is enabled in the deployment.
 
+[id=`configuration-for-distributed-compute-nodes_{context}`]
+=== Configuration for distributed compute nodes
+If you are deploying with compute nodes on different networks to the controller, some additional changes are need so that the metrics sent over AMQ can reach the SAF.
```
leifmadsen: If you are deploying compute nodes on networks different from the controllers, some additional changes are required to allow data transported via AMQ Interconnect to reach the SAF server instance.

docs/README.adoc
```diff
@@ -343,6 +343,20 @@ openstack overcloud deploy \
...
+The following needs to be added under the `ExtraConfig` section for each compute node.
+
+----
```
leifmadsen: Just add this before the `----`

`[source,yaml]`

docs/README.adoc
```diff
@@ -343,6 +343,20 @@ openstack overcloud deploy \
...
+ComputeNExtraConfig:
```
leifmadsen: Is this literally `N` or is it intended to represent a dynamic value?

docs/README.adoc
```diff
@@ -343,6 +343,20 @@ openstack overcloud deploy \
...
+	tripleo::profile::base::metrics::collectd::amqp_host: "%{hiera('internal_apiN')}"
+	tripleo::profile::base::metrics::qdr::listener_addr: "%{hiera('internal_apiN')}"
+----
+
+In the above example, `internal_apiN` is the network that the compute node is connected to.
```
leifmadsen: I'd place this above the example so that when someone is looking at it they understand what items to call out in their head. So I'd reword it a bit like so:

The following configuration example needs to be added under the `ExtraConfig` section for each compute node.

In the example, `internal_apiN` is the network that the compute node is connected to, where `N` represents a value different from the default internal API network name.

```
[source,yaml]
----
code block
----
```

For more information on this configuration, see: https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/13/html-single/spine_leaf_networking/

PR #69, opened 2019-09-30
Adjusted docs for upstream SAF installation

pleimer:
Having the `quickstart.sh` script mentioned in the docs is a little misleading because, unlike what the docs say, it does not take care of all of the steps necessary to launch SAF upstream version. Some of the pre-requisites to running `quickstart.sh` are mentioned later in the docs, causing confusion for the first time viewer. In addition, it is quite simple to quickly launch SAF for development without the script, making it unnecessary.

I also added some additional logic to properly multiplex the 'downstream' and 'upstream' terms.

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -134,6 +134,20 @@ https://github.com/redhat-service-assurance/telemetry-framework/archive/<release
 the host level of each node in your cluster prior to instantiating an SAF
 instance if ElasticSearch is intended to be deployed.
 
+
+ifdef::target-upstream[]
+[id="quickstart-script-for-development-usage_{context}"]
+== Quickstart Script for Development Usage
+
+A script is provided to deploy SAF into an existing
+OpenShift environment. It allows for SAF to be started for development
+purposes, and is not intended for production environments. It takes care of all
+of the required steps for a single-cloud setup. Ensure that Ansible is installed
+before running it.
+
+The script can be found here: https://github.com/redhat-service-assurance/telemetry-framework/blob/master/deploy/quickstart.sh
+endif::[]
+
 ifdef::target-downstream[]
 [id="creating-an-rhcc-secret_{context}"]
 == Creating an RHCC Secret
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -142,8 +142,12 @@ ifdef::target-upstream[]
 A script is provided to deploy SAF into an existing
 OpenShift environment. It allows for SAF to be started for development
 purposes, and is not intended for production environments. It takes care of all
-of the required steps for a single-cloud setup. Ensure that Ansible is installed
-before running it.
+of the required steps for a single-cloud setup. Ensure the following tools are 
+installed before execution:
+
+* Ansible 2.7+
+* OpenShift CLI (oc from https://github.com/openshift/origin/releases/tag/v3.11.0)
+* openssl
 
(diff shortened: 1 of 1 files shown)


csibbitt:
Can you please tell us explicitly which "pre-requisites to running quickstart.sh are mentioned later in the docs"? and which steps "it does not take care of"? It's unclear to me based on this change exactly what you're fixing. 

From the change I would expect the image-streams have something to do with the problem, but quickstart DOES handle image-streams: https://github.com/redhat-service-assurance/telemetry-framework/blob/master/deploy/quickstart.sh#L52

I don't understand the justification for removing the section on the use of quickstart; though I suppose I have no objection to adding the details about running the `./import-upstream.sh` script.

docs/README.adoc
```diff
@@ -134,18 +134,6 @@ https://github.com/redhat-service-assurance/telemetry-framework/archive/<release
 the host level of each node in your cluster prior to instantiating an SAF
 instance if ElasticSearch is intended to be deployed.
 
-ifdef::target-upstream[]
```
csibbitt: :-1: Please be specific about why this should be removed

csibbitt:
Quick additional note on this. I'm fairly certain that quickstart works as designed, because it's used in the smoketest (which failed on this PR due to unrelated reasons): https://github.com/redhat-service-assurance/telemetry-framework/blob/master/.travis.yml#L12

docs/README.adoc
pleimer: I should have been more clear. The `quickstart` script does not handle the [creation of manifests](https://github.com/redhat-service-assurance/telemetry-framework/tree/master/docs#generating-the-manifests-for-saf) now required for the events feature Leif just added. In the documentation, the section mentioning the script comes before the section mentioning the manifests, implying the script handles the creation of these. On a first time install, this will result in the Prometheus operator and every Elasticsearch pod to fail to create because the yaml files do not exist. I ran into this issue when launching SAF on Minishift yesterday. 

If we wanted to keep the script in the docs, the section mentioning it should be moved to the [deployment section](https://github.com/redhat-service-assurance/telemetry-framework/tree/master/docs#deploying-saf-to-the-openshift-environment) of the docs (which again, occurs after mention of the script), and the section adjusted to accommodate using it. I suggested removing it because, once a user reaches this section, the script only automates two of the deployment steps and the user having to do three is not that much bigger of a deal.

pleimer: Upon further investigation it looks like it is supposed to generate the manifests but did not in my case. I will explore further why this was the case

pleimer:
Okay I have gone through the process again, and it looks like `quickstart.sh` does actually work as intended and I just had not installed Ansible on my new machine yet. Maybe that should be mentioned just before encouraging the developer to run the script because it is hard to see that failure in the plethora of output that it generates.

leifmadsen:
We could add a section that lists the prerequisite tooling to execute the script. The script could potentially be updated to check for this as well.

pleimer:
That's a good idea. I can add some checks to the script in a different PR. Tooling is pretty self evident when describing each individual step - maybe it is only necessary for this section?

leifmadsen:
@pleimer yea adding anything to the scripts to do the checks is pretty low priority since hopefully we'll just get away from using all these scripts in the next few weeks (SAF Operator).

However, a simple list of dependent tools for the Quickstart section might be useful. I think the list is something like:

* Ansible 2.7+
* OpenShift CLI (oc from https://github.com/openshift/origin/releases/tag/v3.11.0)
* openssl

PR #72, opened 2019-10-01
Adding telemetry-bench option for performance test

csibbitt:
* telemetry-bench: https://github.com/redhat-service-assurance/telemetry-bench
* Left existing collectd-tg based stuff basically as-is
* Added a very leightweight deployment for telemetry-bench
* Added shell script to abstract between them for easy comparisons
  * Allows command-line driven test params
  * More complex test configs for tg test bench can still be used if needed
* Standardized all default params
  * 20 hosts
  * 1000 plugins (this is roughly realistic)
  * 1 second interval
  * 3 minute test
  * Our testing hardware has us around this limit at the moment, I expect to
    raise these values significantly
* Minimal documentation updates
  * Ultimately I'm not sure which parts of these we will recommend devs use
    or what we'll wire into test automation, so I don't feel like writting any
    throwaway docs on this right now


csibbitt:
[Allow perftest to run with either collectd-tg OR telemetry-bench](https://trello.com/c/M2va4gmK/686-allow-perftest-to-run-with-either-collectd-tg-or-telemetry-bench)

PR #73, opened 2019-10-02
Update to newer golang in perftest container

csibbitt:
https://github.com/grafana-tools/sdk/pull/38/files says we need 1.13

### DIFF
--- a/tests/performance-test/Dockerfile
+++ b/tests/performance-test/Dockerfile
@@ -1,5 +1,5 @@
 #--- Build SAF performance test ---
-FROM golang:1.12.7
+FROM golang:1.13
 WORKDIR /go/src/performance-test/
 
 COPY ./dashboard.go ./main.go ./parser.go ./


csibbitt:
I've actually not been able to verify this due to a package problem in the second stage - reverting this change didn't fix it, so it's independent.

```
Step 7/9 : RUN yum install golang -y &&     yum update -y &&     yum clean all
 ---> Running in a8dc00a5d6a8

--> Finished Dependency Resolution
Error: Package: collectd-rdt-5.8.1-4.el7.x86_64 (@centos-release-opstools)
           Requires: libpqos-1.0.1.so()(64bit)
           Removing: intel-cmt-cat-1.0.1-1.el7.x86_64 (@centos-release-opstools)
               libpqos-1.0.1.so()(64bit)
           Updated By: intel-cmt-cat-3.0.1-1.el7.x86_64 (base)
               Not found
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
The command '/bin/sh -c yum install golang -y &&     yum update -y &&     yum clean all' returned a non-zero code: 1
```

csibbitt:
@pleimer I've assigned this to you because I think you have the only working build process for this container at the moment. You should test this change in whatever workflow you used to get your working image onto quay.io, then merge this to the appropriate place if it works. No rush, just wanted to point out why I haven't merged it yet.

PR #75, opened 2019-10-03
Pull performance test image from quay.io instead of from internal registry

pleimer:
This removes many of the steps required for building the docker image in the performance test by telling the job to pull a static image in Quay.io. We were having a lot of issues building the images locally and this will allow development to continue without the problems of building every time. This is a temporary solution, but necessary for now

### DIFF
--- a/tests/performance-test/deploy/performance-test-job-tg.yml.template
+++ b/tests/performance-test/deploy/performance-test-job-tg.yml.template
@@ -14,7 +14,7 @@ spec:
       restartPolicy: Never
       containers:
         - name: performance-test
-          image: <<REGISTRY_INFO>>/sa-telemetry/performance-test:dev
+          image: quay.io/plmr/performance-test:dev
           imagePullPolicy: Always
           command:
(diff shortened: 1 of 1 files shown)


tests/performance-test/deploy/performance-test-job-tg.yml.template
```diff
@@ -14,7 +14,7 @@ spec:
       restartPolicy: Never
       containers:
         - name: performance-test
-          image: <<REGISTRY_INFO>>/sa-telemetry/performance-test:dev
+          image: quay.io/plmr/performance-test:dev
```
leifmadsen: Do you want access to the quay.io/redhat-service-assurance to host the container image in that namespace instead?

leifmadsen: Actually looks like both Chris and yourself are members there, so should be able to push there. Not saying you have to, but you do have that as an option.

pleimer: Oh yeah I could. Doesn't really matter since this one wont be around long

PR #77, opened 2019-10-08
Fix(?) spurious smoketest failures

csibbitt:
* During CI tests pods can get stuck on registry related errors
  * Usually Unauthorized or Not Found
  * These appear to be race conditions with the setup of the ImageStreams
* In each case, deleting the affected pod leads to recovery
* This "fix" adds a 10m timeout to our rollout checks
  * After timeout, the relevant pod is killed
* In cases where rollout is proceeding smoothly, this add NO extra delay
* In cases where rollouts are failing, delays in multiples of 10m are possible
  * But much higher probability that the rollout will eventually complete

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -143,9 +143,9 @@ EOSM
 if [ "$method" == "CREATE" ]; then
     echo "  * [ii] Creating the operators" ; create "${operator_list[@]}"
     echo "  * [ii] Waiting for prometheus-operator deployment to complete"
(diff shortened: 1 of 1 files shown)


csibbitt:
I'm going to wait for the two :heavy_check_mark: to appear before inviting anyone to the PR ;)

leifmadsen:
So what you're saying is that I shouldn't blindly accept this? :)

csibbitt:
All I'm saying is that when I put up a PR to "fix" the CI tests, the CI tests had better pass or I'm gonna be pretty embarrassed.

PR #78, opened 2019-10-11
Automated performance test with telemetry bench and pre-configured dashboards

pleimer:
This adds in the ability to change automatically run a performance test using telemetry bench instead of collectd-tg. Everything is handled by the `performance-test.sh` script, therefore the process for running a test has changed somewhat. There are two dashboards posted to grafana by the test harness after it completes that gives pre-configured views of OpenShift and SAF metrics during the execution of the test along with several alarms.


tests/performance-test/deploy/performance-test-tg.sh
```diff
@@ -1,21 +1,93 @@
 #!/bin/bash
 # Sets up the envirovnment for performance testing
+
```
csibbitt: I'm not going to comment in this file since most of the same code appears up a level in performance-test.sh and we aren't focused on tg at the moment. I've given it a read-over and it seems sane, but a but out of place with the same logic in the other script. All comments on code in performance-test.sh apply to the same code here of course.

tests/performance-test/performance-test.sh
```diff
@@ -52,28 +95,67 @@ do
...
 fi
+
+
+delete
+get_sources
+create_grafana_datasources
+create_service_monitors
+
+STAGE="TARGET"
+while true; do
```
csibbitt: I'm reviewing a state-machine based operator written in bash. Nice.

tests/performance-test/deploy/scripts/launch-test.sh
```diff
@@ -18,8 +18,7 @@ do
     esac
 done
 
-if [ ! -v LENGTH ];
-then
+if [ ! -v LENGTH ]; then
```
csibbitt: Woah there, cowboy!

tests/performance-test/performance-test.sh
```diff
@@ -1,23 +1,21 @@
 #!/bin/bash
 set -e
 
+
 usage(){
     cat << ENDUSAGE
 Runs on the dev/CI machine to execute a performance test and abstracts between
 running collectd-tg (tg) or telemetry-bench (tb).
-
```
csibbitt: I can see we have stylistic differences. You win this time!

tests/performance-test/performance-test.sh
```diff
@@ -26,17 +24,62 @@ NOTES:
...
   ./performance-test.sh -t tb -c 60 -h 1 -p 1000 -i 1 -n 1
-
   Recommended command for ~20k/s (10 mins)
   ./performance-test.sh -t tb -c 600 -h 5 -p 1000 -i 1 -n 4
 ENDUSAGE
     exit 1
 }
 
+delete(){
+   oc delete servicemonitor/prometheus \
```
csibbitt: :-1: 
I'm not sure I like the idea of deleting the service monitors and waiting for them to re-establish each time we launch some tests. Currently, once my environment is set up, I can execute this script and the test starts up relatively quickly. Now I need to wait for the Prometheus operator to deal with the service monitors each time, and I know from experience that's slow. I run it a few times an hour these days when I'm working on it actively, so it will add up.

performance-test-tb.sh already has `oc delete job/saf-performance-test` in it, so it seems superfluous here. For the service monitors, I think if you change `oc create` to `oc apply` on L50 and L51 below, then you'll find we don't need to delete anything here. Using "apply" is not always safe compared to a delete/create, but I think for our purposes it will be totally fine.

tests/performance-test/performance-test.sh
```diff
@@ -52,28 +95,67 @@ do
...
-    ./performance-test-tg.sh
-
+    echo "Collectd-tg not implemented. Try running with '-t tb' instead"
+    exit 2
 elif [ "${TOOL}" = "tb" ]; then
-    export COUNT HOSTS PLUGINS INTERVAL CONCURRENT
-    cd deploy
-    ./performance-test-tb.sh
-
+    :
```
csibbitt: I'm embarrassed to say this, but I honestly had to look that up.

tests/performance-test/performance-test.sh
```diff
@@ -52,28 +95,67 @@ do
...
+        ;;
+        *)
+            echo "Unrecognized state"
+            exit 2
+        ;;
+    esac
+done
+oc logs -f $(oc get pod -l job-name=saf-perftest-1-runner -o jsonpath='{.items[0].metadata.name}')
+
+# post dashboard
```
csibbitt: This should be done before the tests start (like right after you define the datasources), because I like to watch the dashboard as the test runs. No reason to leave it until after.

tests/performance-test/Dockerfile
```diff
@@ -10,11 +10,11 @@ RUN go get gopkg.in/yaml.v2 && \
     mv main /tmp/
 
 #--- Create performance test layer ---
-FROM tripleomaster/centos-binary-collectd:current-tripleo-rdo
+FROM quay.io/tripleomaster/centos-binary-collectd:current-tripleo
```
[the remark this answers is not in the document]
pleimer: Not actually sure why I changed that, it may have been when I was debugging something.

tests/performance-test/deploy/scripts/launch-test.sh
pleimer: Im all about my coding styles!

tests/performance-test/performance-test.sh
pleimer: I actually prefer the spacing you had in there, it must have been deleted when I resolved conflicts. Adding it back in..

tests/performance-test/deploy/performance-test-tg.sh
pleimer: A lot of the responsibility of this container has been shifted over to the performance-test.sh script, although this configuration does not necessarly show that yet. Once I get around to re-implementing tg, most of this text will disappear

tests/performance-test/performance-test.sh
```diff
@@ -26,17 +24,62 @@ NOTES:
...
+    oc create -f ./grafana/prom-servicemonitor.yml 
+    oc create -f ./grafana/qdr-servicemonitor.yml
+}
+
+get_sources(){
+    oc project openshift-monitoring
+    OCP_PROM_HOST=$(oc get routes --field-selector metadata.name=prometheus-k8s -o jsonpath="{.items[0].spec.host}")
+    OCP_DATASOURCE=$(oc get secret -n openshift-monitoring grafana-datasources -o jsonpath='{.data.prometheus\.yaml}' \
+        | base64 -d)
+    OCP_DATASOURCE=$(echo $OCP_DATASOURCE | python -c "import sys, json; source=json.load(sys.stdin)['datasources'][0]; \
```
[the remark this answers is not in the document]
pleimer: I threw in a regex to get rid of the dependency - I was trying to avoid having to install a tool. Feel free to make a more robust way.

deploy/quickstart.sh
```diff
@@ -7,7 +7,7 @@ set -e
 #  Upstream containers: ./quickstart.sh
 #  Downstream containers: ./quickstart.sh --downstream-secret=~/6340056-cloudops-pull-secret.yaml
 oc login -u system:admin
-oc new-project sa-telemetry
+oc new-project sa-telemetry || true
```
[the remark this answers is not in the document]
pleimer: I just adds a of runtime to the script if the project does already exist. I see what you mean though

tests/performance-test/Dockerfile
```diff
@@ -10,11 +10,11 @@ RUN go get gopkg.in/yaml.v2 && \
...
 USER root
 
 RUN yum install golang -y && \
-    yum update -y && \
+#    yum update -y && \
```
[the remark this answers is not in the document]
pleimer: Would you suggest a link to a new issue or a part of the documentation?

tests/performance-test/performance-test.sh
pleimer: Maybe we should turn it into an actual operator

tests/performance-test/Dockerfile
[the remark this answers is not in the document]
csibbitt: link to https://github.com/redhat-service-assurance/telemetry-framework/issues/81

deploy/quickstart.sh
```diff
@@ -7,6 +7,7 @@ set -e
 #  Upstream containers: ./quickstart.sh
 #  Downstream containers: ./quickstart.sh --downstream-secret=~/6340056-cloudops-pull-secret.yaml
 oc login -u system:admin
+oc delete project sa-telemetry || true
```
csibbitt: Hmmm, now this script is dangerous! 👎 

I don't like this, but I'd like it more if it were behind a flag like `./quickstart.sh --delete` so at least it would have to be deliberate and involve the word 'delete'. See L47 for an example of minimal command line parsing.

It's not significantly better than `oc delete project sa-telemetry; ./quickstart.sh`, though, so I'm not 100% sure I understand the workflow you're trying to solve for with these changes. I **think** the goal is _"I want the ability re-run the script quickly"_ (even quicker than a `oc delete project` if I understood your original comments). My objection to your method (`|| true`) is that it would ignore a failure to create the project, and then start building resources in some other namespace. In the other cases (L21 and L35), it would trigger a failure later on with SSL config which would be harder to debug then a failure here. So I guess what you're really looking for here is a new-project-if-not-exist. You could probably achieve that with `oc get project | grep sa-telemetry` and an if statement. See below for a slicker way to do this for other resources.

deploy/quickstart.sh
```diff
@@ -17,7 +18,7 @@ openssl req -new -x509 -batch -nodes -days 11000 \
         -subj "/O=io.interconnectedcloud/CN=qdr-white.sa-telemetry.svc.cluster.local" \
         -out qdr-server-certs/tls.crt \
         -keyout qdr-server-certs/tls.key
-oc create secret tls qdr-white-cert --cert=qdr-server-certs/tls.crt --key=qdr-server-certs/tls.key
+oc create secret tls qdr-white-cert --cert=qdr-server-certs/tls.crt --key=qdr-server-certs/tls.key || true
```
csibbitt: If you're trying to make this script easy to re-run, then instead of a straight `oc create` here (and on L28), you could try a technique like this: `kubectl create ... --dry-run=true -o yaml | kubectl apply -f -` This lets you construct a new resource with the create command, but still apply it as a patch. It will create the resource if it does not exist, or apply the new config to any existing resource, while leaving any additional config that may have been added intact.

I don't think you can do this with a project, though.

tests/performance-test/Dockerfile
```diff
@@ -14,7 +14,9 @@ FROM tripleomaster/centos-binary-collectd:current-tripleo-rdo
 USER root
 
 RUN yum install golang -y && \
-    yum update -y && \
+    #yum update -y && \  # skipping this step due to issue with
+                         # centos-binary-collectd image. More info:
+                         # https://github.com/redhat-service-assurance/telemetry-framework/issues/81
```
csibbitt: Thanks for this. I'd have been okay with just this line. :) I see the bug even has the error message, so bonus points!

deploy/quickstart.sh
```diff
@@ -31,7 +32,7 @@ oc create secret generic elasticsearch \
     --from-file=logging-es.crt \
     --from-file=admin-key=system.admin.key \
     --from-file=admin-cert=system.admin.crt \
-    --from-file=admin-ca=ca.crt
+    --from-file=admin-ca=ca.crt || true
```
csibbitt: Can't comment down there, but L51 will also need the `oc create ... | oc apply` technique if it is to be re-run. I haven't checked the deploy script itself.

deploy/quickstart.sh
leifmadsen: If you want to speed things up, then perhaps just create an alias to the command?

`alias ocdp='oc delete project'`

pleimer: I will just have it check if the project already exists before having it try to build a new one

deploy/quickstart.sh
```diff
@@ -7,17 +7,26 @@ set -e
 #  Upstream containers: ./quickstart.sh
 #  Downstream containers: ./quickstart.sh --downstream-secret=~/6340056-cloudops-pull-secret.yaml
 oc login -u system:admin
-oc new-project sa-telemetry
+#if [ -z "$(oc projects | grep -o 'sa-telemetry')" ]; then #project does not exist
```
leifmadsen: I don't love comments in files that aren't intending to be added, so hopefully this will go away before merge.

deploy/quickstart.sh
```diff
@@ -7,17 +7,26 @@ set -e
...
+if oc get project sa-telemetry; then
```
leifmadsen: This will fail if the project exists, but you're not already on the `sa-telemetry` project. You need to explicitly change to the `sa-telemetry` project, or at least make sure it's selected.

tests/performance-test/Dockerfile
```diff
@@ -14,7 +14,9 @@ FROM tripleomaster/centos-binary-collectd:current-tripleo-rdo
 USER root
 
 RUN yum install golang -y && \
-    yum update -y && \
+    #yum update -y && \  # skipping this step due to issue with
```
leifmadsen: Is it the full `yum update` that causes an issue? Can you just run the `yum update-minimal --security -y` instead?

I'm not overly worried about this since clearly this is just localized testing, so feel free to skip over.

tests/performance-test/deploy/performance-test-tg.sh
```diff
@@ -1,21 +1,93 @@
...
+SAF_DATASOURCE=$(cat << EOF
+    {
+        "name"  :   "SAFPrometheus",
+        "type"  :   "prometheus",
+        "url"   :   "http://$SAF_PROM_HOST",
+        "access":   "direct",
+        "basicAuth" :false,
+        "jsonData": {
+            "timeInterval"  :   "1s"
```
leifmadsen: Seems like it might be worth to just paramaterize this?

tests/performance-test/deploy/performance-test-tg.sh
```diff
@@ -1,21 +1,93 @@
...
+EOF
+)
+
 TG_CONFIGFILE=${TG_CONFIGFILE:-./config/test-configs.yml}
 
 oc delete configmap/saf-performance-test-collectd-config \
    configmap/saf-performance-test-entrypoint-script \
    configmap/saf-performance-test-configs \
    configmap/saf-performance-test-hosts \
+   configmap/grafana-dashboard-template \
```
leifmadsen: I feel like some of the naming here should be more specific and call out that it's `perfscale-dashboard-template` or something?

leifmadsen: I'm just thinking about future dashboarding work where a generic name like this could potentially cause a conflict.

tests/performance-test/performance-test.sh
```diff
@@ -1,15 +1,15 @@
 #!/bin/bash
 set -e
 
+
```
leifmadsen: A double blank line?! :astonished:

tests/performance-test/performance-test.sh
```diff
@@ -52,28 +124,59 @@ do
...
-    plugins: ${PLUGINS}
-    interval: ${INTERVAL}
-    length: ${LENGTH_S}"
-
-    echo "${CONFIG}" > deploy/config/generated-test-configs.yml
-    cd deploy
-    export TG_CONFIGFILE=./config/generated-test-configs.yml
-    ./performance-test-tg.sh
-
+    echo "Collectd-tg not implemented. Try running with '-t tb' instead"
```
leifmadsen: :+1:

tests/performance-test/performance-test.sh
```diff
@@ -52,28 +124,59 @@ do
...
+get_sources
+create_grafana_datasources
+create_service_monitors
+post_dashboards
+
+STAGE="TARGET"
+while true; do
+    case $STAGE in
+        "TARGET")
+            TARGETS=`curl "http://$(oc get route prometheus -o jsonpath='{.spec.host}')/api/v1/targets"`
```
leifmadsen: Is this the default Prometheus path that is created by the `deploy.sh`? I would expect this to fail at some point in the future since it's not using an `https` endpoint. Not sure if the `https` endpoint exists right now, so it's entirely possible this is the currently valid approach.

Future you could use `curl -k https://...` I think.

tests/performance-test/performance-test.sh
```diff
@@ -52,28 +124,59 @@ do
...
+            if [ "${estab}" != "True" ]; then
+                echo "Waiting on SAF performance test pod creation..."
+                sleep 3
+            else
+                echo "SAF performance test pod established"
+                break
+            fi
+        ;;
+        *)
+            echo "Unrecognized state"
```
leifmadsen: Would it be appropriate to call `usage()` here?

deploy/quickstart.sh
pleimer: Left it in there on accident, I will make another sweep open merging

deploy/quickstart.sh
pleimer: My more recent commits take care of this

tests/performance-test/deploy/performance-test-tg.sh
pleimer: Valid point. It should be more uniquely identifiable for its use case

tests/performance-test/performance-test.sh
pleimer: It currently is valid because the Prometheus in SAF does not use https

tests/performance-test/performance-test.sh
pleimer: Potentially, though theoretically control flow should never arrive here and will definitely never arrive here because of user error. Its just a default to keep anything unexpected from happening if something like radiation happened to flip the bits in the state variable

tests/performance-test/deploy/performance-test-tg.sh
pleimer: For collectd-tg yes. In fact, most of this script's functionality is now apart of the performance-test.sh script and it should be stripped down. Priority shifted towards getting everything working with telemetry-bench first, therefore a lot of the collectd-tg is midway through re factoring and probably does not work at all anyway. Really, tg stuff is outside the scope of this feature request and should be dealt with separately

deploy/quickstart.sh
pleimer: That is a good point

PR #83, opened 2019-10-22
Configured edge router to accept connection from telemetry bench and …

pleimer:
In order to more realisticly test SAF, the setup needed to more closely reflect a real life setup in which data would come into an edge router on the client side before being routed over to the amq bus in SAF: tool -> qdr -> qdr -> sg.

I have added in a new router to satisfy this condition. The performance test automatically deploys and tracks the new router and points telemetry bench to send metrics through it.

### DIFF
--- a/tests/performance-test/deploy/performance-test-job-tb.yml.template
+++ b/tests/performance-test/deploy/performance-test-job-tb.yml.template
@@ -17,7 +17,7 @@ spec:
         - name: performance-test
           image: quay.io/redhat-service-assurance/telemetry-bench
           imagePullPolicy: Always
-          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-white.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
+          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-test.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
       affinity:
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
--- a/tests/performance-test/deploy/qdrouterd.yaml
+++ b/tests/performance-test/deploy/qdrouterd.yaml
@@ -20,7 +20,7 @@ spec:
       distribution: closest
     - prefix: broadcast
       distribution: multicast
-    - prefix: collectd
+    - prefix: collectd/telemetry
       distribution: multicast
     - prefix: collectd/notify
       distribution: multicast
--- a/tests/performance-test/deploy/qdrouterd.yaml
+++ b/tests/performance-test/deploy/qdrouterd.yaml
@@ -1,0 +1,42 @@
+apiVersion: interconnectedcloud.github.io/v1alpha1
+kind: Qdr
+metadata:
+  name: qdr-test
+  namespace: sa-telemetry
+spec:
+  deploymentPlan:
+      image: docker-registry.default.svc:5000/sa-telemetry/amq-interconnect:1.4-7
+      role: edge
+      size: 1
+      placement: Any
+  addresses:
+    - prefix: closest
+      distribution: closest
+    - prefix: multicast
+      distribution: multicast
+    - prefix: unicast
+      distribution: closest
+    - prefix: exclusive
+      distribution: closest
+    - prefix: broadcast
+      distribution: multicast
+    - prefix: collectd
+      distribution: multicast
+    - prefix: collectd/notify
+      distribution: multicast
+  sslProfiles:
+    - name: default
+      credentials: "qdr-white-cert"
+  edgeListeners:
+    - host: "0.0.0.0"
+      port: 5671
+      expose: true
+  listeners:
+    - port: 5672
+    - port: 8672
+      http: true
+  edgeConnectors:
+    - name: router
+      host: qdr-white.sa-telemetry.svc.cluster.local
+      port: 5671
+      verifyHostname: no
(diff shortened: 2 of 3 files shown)


tests/performance-test/deploy/performance-test-job-tb.yml.template
```diff
@@ -17,7 +17,7 @@ spec:
         - name: performance-test
           image: quay.io/redhat-service-assurance/telemetry-bench
           imagePullPolicy: Always
-          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-white.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
+          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-test.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
```
leifmadsen: nice, this is a better naming scheme

tests/performance-test/deploy/qdrouterd.yaml
```diff
@@ -0,0 +1,42 @@
...
+      distribution: closest
+    - prefix: multicast
+      distribution: multicast
+    - prefix: unicast
+      distribution: closest
+    - prefix: exclusive
+      distribution: closest
+    - prefix: broadcast
+      distribution: multicast
+    - prefix: collectd
```
leifmadsen: wouldn't this be collectd/telemetry?

tests/performance-test/deploy/qdrouterd.yaml
```diff
@@ -0,0 +1,42 @@
...
+      distribution: multicast
+    - prefix: collectd/notify
+      distribution: multicast
+  sslProfiles:
```
leifmadsen: you don't need to specify a cert here, it should just accept what the interior router is using much like we do on the client side for openstack. There is just an `sslProfile` with a `name: default` and no certificates loaded. Then the connection is setup via SSL, but is implicitly authenticated.

leifmadsen: I'm not exactly sure how all of that works in the Operator based deployment though, so it's possible that configuration for OpenShift isn't valid. Would need to check, but I suppose since you have access to the cert anyways, then why not?

In the future hopefully we'll get all that security stuff sorted out.

tests/performance-test/performance-test.sh
```diff
@@ -35,17 +35,32 @@ ENDUSAGE
     exit 1
 }
 
+ELIPSE="."
```
leifmadsen: Can you add some docs on what this function in providing? I see that it changes one type of ellipses (note the spelling, should be 2 L's) to another, but it's not exactly clear at this point why this is necessary.

tests/performance-test/performance-test.sh
```diff
@@ -58,6 +73,17 @@ create_service_monitors(){
     oc apply -f ./grafana/qdr-servicemonitor.yml
 }
 
+make_qdr_edge_router(){
```
leifmadsen: `make_qdr_edge_router_great_again()`

tests/performance-test/performance-test.sh
```diff
@@ -141,28 +167,40 @@ while true; do
...
-                echo "Found new target endpoints. Creating performance test job"
+                echo "Found new target endpoints"
+                STAGE="ROUTER"
+            fi
+        ;;
+        "ROUTER")
+            make_qdr_edge_router
+            estab=$(oc get pod -l qdr_cr=qdr-test -o jsonpath='{.items[0].status.conditions[?(@.type=="ContainersReady")].status}') || true
+            if [ "${estab}" != "True" ]; then
+                printf "%s" "Waiting on qdr edge test pod creation"; elipses
```
leifmadsen: s/elipses/ellipses/g

tests/performance-test/performance-test.sh
```diff
@@ -141,28 +167,40 @@ while true; do
...
+            else
+                echo "Qdr edge test pod established. Creating performance test job"
                 export COUNT HOSTS PLUGINS INTERVAL CONCURRENT
                 cd deploy
                 ./performance-test-tb.sh
                 STAGE="TEST"
             fi
         ;;
-        "TEST")
+        "TEST")
```
leifmadsen: trailing space

tests/performance-test/deploy/qdrouterd.yaml
pleimer: This I directly took from your config for the SAF router, so I figured I shouldn't change it?

tests/performance-test/deploy/qdrouterd.yaml
pleimer: You're right. I used this when I was trying to debug something before I understood all that the operator was doing

tests/performance-test/deploy/qdrouterd.yaml
leifmadsen: Oh really? Hrmmm... something seems off here then. I think Chris updated this to `collectd/telemetry` and `collectd/notify` for the metrics vs events addresses.

For testing purposes, whatever works I suppose, but I would have expected this to be `telemetry` not just `collectd`. I could be misunderstanding though.

tests/performance-test/deploy/qdrouterd.yaml
pleimer: It's asymmetrical authentication so there isn't need for that here, just remnant code I accidently left in

tests/performance-test/performance-test.sh
pleimer: For sure. Really just to make it clear that the script is not frozen when it is waiting on pods to deploy. If an image must be pulled from an online repository, it can take a while

tests/performance-test/performance-test.sh
pleimer: oops

tests/performance-test/deploy/qdrouterd.yaml
pleimer: It works either way so I will go ahead and change it so we don't get unintended messages skewing our data later on

csibbitt:
[Make more realistic test tool setup (tool -> qdr -> qdr -> sg)](https://trello.com/c/XMZ5et62/682-make-more-realistic-test-tool-setup-tool-qdr-qdr-sg)

PR #86, opened 2019-10-29
Simplified collectd prefix in qdr config

csibbitt:
Paul had this right initially in the PR. This shorter prefix is simpler, more flexible, and more closely matches what the client side uses for real. See: https://github.com/openstack/tripleo-heat-templates/commit/13ffaf688ecabb02378e8ef31107ae2be7fd1010


csibbitt:
I saw a comment in the original PR thread

> It works either way so I will go ahead and change it so we don't get unintended messages skewing our data later on

Just to clarify this point - this config shouldn't affect anything about how stats are collected for specific addresses in the qdr, just how it routes messages, so I don't think this will cause any complication in stats measurements.

PR #91, opened 2019-11-05
Events verifiable in Performance Test

pleimer:
Breaking all of the rules, there are a few things done in this PR:

1) Event verification added to the performance test script and Grafana dashboard. Currently there is only a test to verify that events made it from collectd, through the QDR routers and into Elastic Search. It is fairly convoluted how this process happens: the PF script gathers event logs from the collectd pod, queries Elastic Search to see what events were received by the most recent test, and builds a document to post to Elastic Search for Grafana to find. The dashboard only displays event results for the most recently run test. Once a new test is run, previous event results are erased. 

2) Data source creation was moved to the provisioning stage of Grafana instead of being handled by the performance test script. This allows for cleaner, more reliable code and dashboards. When running the test, ensure that the 'grafana-launcher.sh' script is run before to ensure your environment has the most up to date Grafana configuration (otherwise events will not work!)

3) Cleaned up test output


tests/performance-test/deploy/config/minimal-collectd.conf
```diff
@@ -1,23 +1,41 @@
...
-LoadPlugin network
-
 <Plugin "amqp1">
   <Transport "name">
-    Host "qdr-white.sa-telemetry.svc.cluster.local"
+    Host "qdr-test.sa-telemetry.svc.cluster.local"
     Port "5672"
     Address "collectd"
-    <Instance "telemetry">
+    <Instance "notify">
```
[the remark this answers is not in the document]
pleimer: Well I use telemetry bench for metrics testing. Since its an entirely different pod this shouldn't matter

tests/performance-test/grafana/datasource.yaml
```diff
@@ -0,0 +1,62 @@
...
+  type: prometheus
+  access: proxy
+  orgId: 1
+  url: "http://prometheus-operated.sa-telemetry.svc:9090"
+  database: ""
+  basicAuth: false
+  withCredentials: false
+  isDefault: false
+  jsonData:
+    timeInterval: "1s"
```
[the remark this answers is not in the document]
csibbitt: This sets the lower limit on the "step" parameter for queries, which protects you from misreading your silly things when your queries demand a higher resolution than prometheus actually has. Having too low a value also causes higher CPU demand during prometheus queries as it fills in data from further back in the db. See this issue[1] for a good set of hints on this value and the potential problems.

Ideally it should be set to whatever the prometheus scrape interval is set to (see: https://github.com/redhat-service-assurance/smart-gateway-operator/pull/19) The easiest fix is probably to query this from Prometheus and re-use the value in a datasource template, but it's somewhat tricky because it can actually have different scrape intervals for different targets; choosing the global default might be best? 

This isn't too critical for our purposes so I wouldn't hold the PR up for it unless we observe it doing bad things. When grafana becomes part of SAF proper I think this will actually be easier to solve.

[1] https://github.com/prometheus/prometheus/issues/2564

tests/performance-test/deploy/config/minimal-collectd.conf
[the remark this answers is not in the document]
csibbitt: Neat to see we're finding a use for this code at least! :)

tests/performance-test/grafana/datasource.yaml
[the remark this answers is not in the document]
csibbitt: Whoops, this comment belonged on the prometheus datasource. Not sure what to set it to for the ES datasource. 1s? ;)

tests/performance-test/grafana/grafana-deploy.yml
```diff
@@ -19,13 +19,36 @@ spec:
...
+           - name: CLIENTKEY
+             valueFrom:
+              secretKeyRef:
+                name: elasticsearch
+                key: admin-key
          volumeMounts:
            - name: grafana-config
              mountPath: /etc/grafana/grafana.ini
              subPath: grafana.ini
+           - name: datasources
```
csibbitt: :+1: You told me about this the other day. Much cleaner, cool.

tests/performance-test/performance-test.sh
```diff
@@ -49,91 +49,170 @@ ellipse(){
...
-    if ! oc project sa-telemetry; then 
+# basic check if necessary resources exist. DOES NOT verify they work correctly
+check_resources(){
+    printf "\n*** Performing Resource Checks ***\n"
+    if ! oc get project sa-telemetry; then 
         echo "Error: SAF does not exist on cluster. Deploy SAF before running performance test" 1>&2
         exit 1
     fi
+
+    if ! oc get project openshift-monitoring; then
```
csibbitt: Nice!

tests/performance-test/performance-test.sh
```diff
@@ -49,91 +49,170 @@ ellipse(){
...
     curl --silent --output /dev/null -d "{\"overwrite\": true, \"dashboard\": $(cat ./grafana/prom2-dashboard.json)}" \
         -H 'Content-Type: application/json' "$GRAF_HOST/api/dashboards/db"
 }
 
-# Execute
+# Delete recorded events in Elastic Search
+delete_es_events(){
+    local esPod=$(oc get elasticsearch elasticsearch -o jsonpath='{.status.pods.master.ready[0]}')
+    local secretPath="/etc/elasticsearch/secret/"
+    local esQuery=$(cat <<EOF
```
csibbitt: Just a personal preference, but I'd put all this in one wide line, or at LEAST collapse all the `}}}}}` into one line at the end. I bet others might disagree :)

tests/performance-test/performance-test.sh
```diff
@@ -49,91 +49,170 @@ ellipse(){
...
+                        }
+                    }
+                }
+            }
+        }
+EOF
+)
+    printf "\n*** Deleting ES Events ***\n"
+
+    local response=$(oc exec "$esPod" -- curl -k --silent --trace --cacert "${secretPath}admin-ca" --cert \
```
csibbitt: `oc exec "$esPod" -- curl -k --silent --trace....` appears 4 times so deserves to be it's own function. Not worth blocking on, but something to note for a potential future refactor.

tests/performance-test/performance-test.sh
```diff
@@ -49,91 +49,170 @@ ellipse(){
...
+                }
+            }
+        }
+EOF
+)
+    ES_EVENT_RECV_COUNT=$(oc exec "$esPod" -- curl -k --silent --trace --cacert "${secretPath}admin-ca" --cert \
+         "${secretPath}admin-cert" --key "${secretPath}admin-key" -X GET "https://localhost:9200/collectd_interface_if/_count" \
+         -d "$esQuery")
+
+    ES_EVENT_RECV_COUNT=$(echo "$ES_EVENT_RECV_COUNT" | python -c "import sys, json; print json.load(sys.stdin)['count']")
```
csibbitt: Bringing back this python dependency I see. We managed to drop it previously when we stopped using performance-test-tg.sh. I won't harp on it since I know you wanted to re-write the whole thing in python anyways ;)

tests/performance-test/performance-test.sh
```diff
@@ -49,91 +49,170 @@ ellipse(){
...
+        exit 1
+    fi
+
+    if ! echo $datasources | grep -q "SAFElasticsearch"; then
+        echo "Error: unable to find Grafana datasource SAFElasticsearch"
+        exit 1
+    fi
 }
 
+# Post dashboards to grafana - overwrites if they already exist
```
csibbitt: Just a comment that I actually nerf this every time because I tend to tweak dashboards as I'm testing. An option to disable it would be nice.

tests/performance-test/performance-test.sh
```diff
@@ -158,37 +237,41 @@ else
...
             QDR=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"qdr-white"')
             PROM=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"prometheus-operated"')
 
             if [ -z "$QDR" ] && [ -z "$PROM" ]; then
                 printf "%s" "Waiting for new targets to be recognized by Prometheus Operator"; ellipse
-                printf "\r"
                 sleep 10
             else
                 echo "Found new target endpoints"
+                make_qdr_edge_router
```
csibbitt: :question: I'm interested to know why this moved?

tests/performance-test/performance-test.sh
```diff
@@ -49,91 +49,170 @@ ellipse(){
...
     oc apply -f ./grafana/qdr-servicemonitor.yml
 }
 
+# create edge router for more realistic env setup: telemetry-bench -> qdr -> qdr -> SG
 make_qdr_edge_router(){
     if ! oc get qdr qdr-test; then
-        echo "Deploying edge router"
-        oc create -f <(sed -e "s/<<REGISTRY_INFO>>/$(oc registry info)/" ./deploy/qdrouterd.yaml.template)
+        printf "\n*** Deploying Edge Router ***\n"
+        sed -e "s/<<REGISTRY_INFO>>/$(oc registry info)/g" ./deploy/qdrouterd.yaml > /tmp/qdr-test.yaml
```
csibbitt: Don't like my one-liner? Pulled it apart to find the missing `g` and never put it back together? ;) Not offended.

tests/performance-test/deploy/config/minimal-collectd.conf
```diff
@@ -1,23 +1,41 @@
 Interval 1
 
+LoadPlugin "log_logstash"
+<Plugin "log_logstash">
+  LogLevel "notice"
+  File "/tmp/events.json"
```
csibbitt: I was wondering what this was about, and it took almost until the end of the story for it to show itself. Clever!

tests/performance-test/performance-test.sh
```diff
@@ -199,22 +282,47 @@ while true; do
...
+                oc logs -f "$(oc get pod -l job-name=saf-perftest-1-runner -o jsonpath='{.items[0].metadata.name}')" |  grep -E 'total [0-9]+'
+                STAGE="RESULTS"
             fi
         ;;
+        "RESULTS")
+            printf "\n*** Collecting test results ***\n"
+            PODNAME=$(oc get pod -l job-name=saf-perftest-notify -o jsonpath='{.items[0].metadata.name}')
+            oc exec "$PODNAME" -- pkill collectd > /dev/null
+            sleep 3
+            oc cp "sa-telemetry/${PODNAME}:/tmp/events.json" /tmp/events.json
```
csibbitt: :+1:

tests/performance-test/deploy/performance-test-tb.sh
```diff
@@ -8,12 +8,23 @@ CONCURRENT=${CONCURRENT:-1}
...
-  oc create -f <(sed  -e "s/<<PREFIX>>/saf-perftest-${i}-/g;
-                          s/<<COUNT>>/${COUNT}/g;
-                          s/<<HOSTS>>/${HOSTS}/g;
-                          s/<<PLUGINS>>/${PLUGINS}/g;
-                          s/<<INTERVAL>>/${INTERVAL}/g"\
-                          performance-test-job-tb.yml.template)
+oc create configmap saf-performance-test-events-entry --from-file \
+          ./entrypoint.sh
+          
+oc create -f ./performance-test-job-events.yml.template
```
csibbitt: This is a bit weird, historically speaking, having `performance-test-tb.sh` calling collectd-tg. Since you deleted the tg version of this script, then.... okay sure :)

tests/performance-test/deploy/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+/usr/sbin/collectd -C /tmp/minimal-collectd.conf -f 2>&1 | tee /tmp/collectd_output
+
+sleep 5
```
csibbitt: I came back to this PR to answer the questions in my mind "How does the events pod get shut down?" because I missed it the first time through. I see that this sleep co-ordinates with a `sleep 3` below while you wait for collectd to shut down then copy the file out before the pod terminates. Tricky! This is probably safe enough for our purposes but try to keep co-ordinate-by-sleep out of anything that runs in production.

tests/performance-test/performance-test.sh
pleimer: I should add datasources to the Grafana provisioning stage anyways - then it will only be written once. Future PR

tests/performance-test/performance-test.sh
pleimer: I think the second thing is what happened. I like the one liner!

tests/performance-test/deploy/performance-test-tb.sh
pleimer: Agreed it is a little strange. Calling it 'notify' instead of collectd may keep that mental clash from happening

tests/performance-test/performance-test.sh
pleimer: I think I gave up after the fourth time trying to parse JSON with regular expressions

tests/performance-test/grafana/grafana-deploy.yml
leifmadsen: Ooooh nice!

tests/performance-test/performance-test.sh
leifmadsen: I guess you could technically put it back to the one-liner since you use that same method in other places throughout?

tests/performance-test/deploy/entrypoint.sh
pleimer: For sure. The problem here is that this script is limited to using the 'oc' tool which doesn't give all of the controller ability that a native python or golang api provides, so I had to resort to some hacks

tests/performance-test/performance-test.sh
leifmadsen: Don't forget to open a low priority issue so you don't forget! :)

tests/performance-test/performance-test.sh
leifmadsen: I'd disagree :)

pleimer: That's an interesting way to do it

tests/performance-test/performance-test.sh
pleimer: How would the function account for different arguments and queries? For example some of them I have to POST while others are a DELETE

tests/performance-test/performance-test.sh
pleimer: I believe it may be wrong in the master branch - if the edge routers are not established already, then the loop will continuously tell oc to build the monitors, even if oc is already trying to

tests/performance-test/performance-test.sh
pleimer: Good point!

tests/performance-test/performance-test.sh
pleimer: Yes I definitely can

PR #94, opened 2019-11-11
Set telemetry-bench to send unsettled messages in Perf Test

pleimer:
After some investigation as to why the QDR->QDR setup in the Performance Test was causing messages to be dropped by qdr, I set telemetry bench to send unsettled messages as opposed to pre-settled. After some testing, this seems to have solved the issue.

With pre-settled messages: 

image

With unsettled messages:

image

### DIFF
--- a/tests/performance-test/deploy/entrypoint.sh
+++ b/tests/performance-test/deploy/entrypoint.sh
@@ -2,4 +2,4 @@
 
 /usr/sbin/collectd -C /tmp/minimal-collectd.conf -f 2>&1 | tee /tmp/collectd_output
 
-sleep 5
+sleep 10
--- a/tests/performance-test/deploy/performance-test-job-tb.yml.template
+++ b/tests/performance-test/deploy/performance-test-job-tb.yml.template
@@ -17,7 +17,7 @@ spec:
         - name: performance-test
           image: quay.io/redhat-service-assurance/telemetry-bench
           imagePullPolicy: Always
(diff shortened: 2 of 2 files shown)


pleimer:
Peak messages throughput does go down, however. 

Test configuration for both: ./performance-test.sh -t tb -c 60 -h 1 -p 3000 -i 1 -n 1

csibbitt:
I'm going to need to test this on the big hardware to be sure, but I don't think this solves the problem I pointed out:

> I ran ./performance-test.sh -t tb -c 360 -h 10 -p 1000 -i 10 -n 10 (Recommend a lower -c if you want to reproduce) and found that only 19929542 messages were received by the QDR instead of the expected 36000000. Pretty sure this was not happening before. [It wasn't]

The test I showed generates 10k metrics/s and the complaint is that the "Received by the QDR" was too low, showing that some messages were dropped either between T-B and the edge QDR, or between the edge QDR and the SAF QDR. 

The output you show does not reproduce the problem "Received by the QDR" appears _possibly_ correct in both cases (If it was the same command I'd expect the EXACT same number though, so something is fishy here). You appear to be declaring success due because the "Dropped by QDR" has hit zero, but this represents drops between the SAF QDR and the SG, not the problem I saw at all. Requiring ACKs is definitely supposed to improve the end-to-end reliability, but it's going to cost  you in peak throughput barely reaching 1600 metrics/s (on what I assume was a 3k metrics/s test based on the numbers).

In either case, we don't want unsettled metrics, so I don't want to merge this patch. I will work towards re-testing my scenario to prove whether or not this changes what I observed.

csibbitt:
Closing this PR, opened this issue: https://github.com/redhat-service-assurance/telemetry-framework/issues/96

PR #97, opened 2019-11-13
Eliminate 'bc' dependency from perftest script

csibbitt:
You don't know how lucky you are growing up in an era of fast floating point arithmetic! :smile: Shifting the decimal like this is a good tool to have in your pocket.

The test rig I've been using for perf testing didn't have `bc` installed. No biggie, but figured we could skip a `yum install` of this in our future automation. We should aggressively limit dependencies until all of this is containerized (if ever).

### DIFF
--- a/tests/performance-test/performance-test.sh
+++ b/tests/performance-test/performance-test.sh
@@ -304,7 +304,7 @@ while true; do
             get_es_event_count
 
             # calulate success ratio
-            SUCCESS_RATIO=$(echo "scale=2; $ES_EVENT_RECV_COUNT / $NUM_COLLECTD_EVENTS" | bc -l | awk '{printf "%.2f", $0}')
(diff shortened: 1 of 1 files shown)


csibbitt:
# Testing
I haven't actually run the script because I'm working on something else, but I've got this:
```
[csibbitt@laptop telemetry-framework (master) ]$ (
> ES_EVENT_RECV_COUNT=4
> NUM_COLLECTD_EVENTS=9
> echo -n "OLD VERSION: "
> echo $(echo "scale=2; $ES_EVENT_RECV_COUNT / $NUM_COLLECTD_EVENTS" | bc -l | awk '{printf "%.2f", $0}')
> echo -n "NEW VERSION: "
> echo $(( ES_EVENT_RECV_COUNT / NUM_COLLECTD_EVENTS)).$(( (ES_EVENT_RECV_COUNT * 100 / NUM_COLLECTD_EVENTS) % 100 ))
> )
OLD VERSION: 0.44
NEW VERSION: 0.44
```

leifmadsen:
Quick test for bot. Sorry, will delete after if I can.

leifmadsen:
Quick test. Will delete.

PR #99, opened 2019-11-14
Edge router monitoring added to the perftest dashboard

pleimer:
I also put in an events latency panel as a place holder, but it does not actually work yet. Feel free to make some layout recommendations you may think look better.

Edge router results work with all currently implemented alarms. Separate qdr-test metrics are graphed in the 'QDR' row of the dashboard (not shown below). This is a fairly un-intrusive update and will work even if the test is run with metrics sent directly to qdr-white.

**Dashboard View**
image

### DIFF
--- a/tests/performance-test/deploy/performance-test-job-tb.yml.template
+++ b/tests/performance-test/deploy/performance-test-job-tb.yml.template
@@ -17,7 +17,7 @@ spec:
         - name: performance-test
           image: quay.io/redhat-service-assurance/telemetry-bench
           imagePullPolicy: Always
-          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-test.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
+          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-white.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
       affinity:
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
--- a/tests/performance-test/grafana/qdr-servicemonitor.yml
+++ b/tests/performance-test/grafana/qdr-servicemonitor.yml
@@ -1,13 +1,28 @@
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: 'qdr'
+  name: 'qdr-white'
   labels:
     app: smart-gateway
 spec:
   selector:
     matchLabels:
       application: qdr-white
+  endpoints:
+  - port: port-8672
+    interval: 10s 
+
+---
+apiVersion: monitoring.coreos.com/v1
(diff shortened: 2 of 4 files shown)


pleimer:
When I get latency fully inplemented, there will most likely be some additional panels to add as well.

tests/performance-test/grafana/qdr-servicemonitor.yml
```diff
@@ -11,3 +11,18 @@ spec:
   endpoints:
   - port: port-8672
     interval: 10s 
+
+---
```
csibbitt: I don't like that we are adding a servicemonitor for each QDR service; ideally there should be a label that can identify all of them. Unfortunately I checked and there is not; not is is immediately obvious to me how we'd get a label added to those services. This will have to do for now, thanks!

tests/performance-test/performance-test.sh
```diff
@@ -251,10 +251,11 @@ while true; do
     case $STAGE in
         "TARGET")
             TARGETS=$(curl --silent "http://$(oc get route prometheus -o jsonpath='{.spec.host}')/api/v1/targets")
-            QDR=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"qdr-white"')
+            QDRWHITE=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"qdr-white"')
+            # QDRTEST=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"qdr-test"')
```
csibbitt: Is there some reason you're not looking for both here?

pleimer: If the new qdr-test has not been completely established yet, the test exits without any indication of what happened. I need to add a section in to wait on the qdr-test instantiation. I thought I would do it in a separate PR but I think I will go ahead and do it here

tests/performance-test/deploy/performance-test-job-tb.yml.template
```diff
@@ -17,7 +17,7 @@ spec:
         - name: performance-test
           image: quay.io/redhat-service-assurance/telemetry-bench
           imagePullPolicy: Always
-          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-test.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
+          args: ["-hostprefix", "<<PREFIX>>", "-hosts", "<<HOSTS>>", "-plugins", "<<PLUGINS>>", "-instances", "1", "-send", "<<COUNT>>", "-interval", "<<INTERVAL>>", "-startmetricenable", "-verbose", "amqp://qdr-white.sa-telemetry.svc.cluster.local:5672/collectd/telemetry/"]
```
csibbitt: This looks to be outside the scope of this PR, but I'm okay with either value, so sure.

tests/performance-test/performance-test.sh
```diff
@@ -126,24 +139,12 @@ delete_es_events(){
...
-                }
-            }
-        }
+        { "query" : { "bool": { "must": { "match_phrase": { "labels.instance": { "query": "saf-perftest-notify-*" } } } } } }
 EOF
 )
     printf "\n*** Deleting ES Events ***\n"
 
-    local response=$(oc exec "$esPod" -- curl -k --silent --trace --cacert "${secretPath}admin-ca" --cert \
+    local response=$(oc exec "$esPod" -- curl --silent --cacert "${secretPath}admin-ca" --cert \
```
csibbitt: :+1: When I noticed this it was because I saw a file called `--cacert` on my ES pod.

pleimer: Nice catch!

PR #100, opened 2019-11-19
New supportable defaults for integration testing

csibbitt:
* Setting these as reasonable defaults for integration testing
* Note the change from CollectdExtraPlugins to CollectdDefaultPlugins
  * This effectively overrides this list: https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/metrics/collectd-container-puppet.yaml#L80
  * Adds: intel_rdt, ipmi, memory, ovs_events, ovs_stats
  * Removes: tcpconns
* This config produces <100 metrics/s/node
  * Down from ~1000 with the previous version of this file
  * Allows us the possibility of meeting the 100 nodes/cloud requirement
  * \# of VMs and network interfaces have the largest impact in production
* I wanted to explicitly set the interval of every plugin
  * Found that collectd::plugin::ovs_stats::interval does not work/exist
  * Adjusted overall interval to 5s as that's the lowest we use

**NOTE**: I really want to include the `virt` plugin here (It is included in the testing below) but I got triple-o errors when trying to deploy it to a non-compute node. I could use some help with the correct THT role implementation to selectively enable it here and upstream!

### DIFF
--- a/tests/infrared/metrics-collectd-qdr.yaml.template
+++ b/tests/infrared/metrics-collectd-qdr.yaml.template
@@ -16,8 +16,6 @@ custom_templates:
             - ipmi
             - load
             - memory
-            - ovs_events
-            - ovs_stats
             - processes
             - unixsock
             - uptime
--- a/tests/infrared/metrics-collectd-qdr.yaml.template
+++ b/tests/infrared/metrics-collectd-qdr.yaml.template
@@ -4,28 +4,47 @@ tripleo_heat_templates:
 
 custom_templates:
     parameter_defaults:
-        CollectdAmqpInterval: 1
-        CollectdDefaultPollingInterval: 1
-        CollectdExtraPlugins:
+        CollectdAmqpInterval: 5
+        CollectdDefaultPollingInterval: 5
+        CollectdDefaultPlugins:
             - cpu
             - df
+            - disk
             - hugepages
+            - intel_rdt
+            - interface
+            - ipmi
+            - load
+            - memory
             - ovs_events
             - ovs_stats
-            - load
+            - processes
+            - unixsock
             - uptime
         DnsServers: ["10.19.42.41","10.11.5.19","10.5.30.160"]
         ExtraConfig:
-            collectd::plugin::cpu::valuespercentage: true
+            collectd::plugin::cpu::interval: 5
             collectd::plugin::cpu::reportbycpu: true
             collectd::plugin::cpu::reportbystate: true
             collectd::plugin::cpu::reportnumcpu: false
+            collectd::plugin::cpu::valuespercentage: true
+            collectd::plugin::df::interval: 300
             collectd::plugin::df::ignoreselected: true
+            collectd::plugin::df::fstypes: ['xfs']
+            collectd::plugin::disk::interval: 5
+            collectd::plugin::hugepages::interval: 60
+            collectd::plugin::intel_rdt::interval: 5
+            collectd::plugin::interface::interval: 5
+            collectd::plugin::ipmi::interval: 60
+            collectd::plugin::load::interval: 60
             collectd::plugin::load::reportrelative: true
-            collectd::plugin::interface::ignoreselected: true
+            collectd::plugin::memory::interval: 5
+            collectd::plugin::processes::interval: 5
+            collectd::plugin::uptime::interval: 60
+            collectd::plugin::virt::interval: 5
             collectd::plugin::virt::connection: "qemu:///system"
-            collectd::plugin::virt::hostname_format: "hostname uuid"
             collectd::plugin::virt::extra_stats: "cpu_util disk disk_err pcpu job_stats_background perf vcpupin"
+            collectd::plugin::virt::hostname_format: "hostname uuid"
         MetricsQdrAddresses:
             - prefix: 'collectd'
               distribution: multicast


tests/infrared/metrics-collectd-qdr.yaml.template
```diff
@@ -4,28 +4,47 @@ tripleo_heat_templates:
...
         ExtraConfig:
-            collectd::plugin::cpu::valuespercentage: true
+            collectd::plugin::cpu::interval: 5
             collectd::plugin::cpu::reportbycpu: true
             collectd::plugin::cpu::reportbystate: true
             collectd::plugin::cpu::reportnumcpu: false
+            collectd::plugin::cpu::valuespercentage: true
+            collectd::plugin::df::interval: 300
             collectd::plugin::df::ignoreselected: true
+            collectd::plugin::df::fstypes: ['xfs']
```
csibbitt: This ignores all of the bind-mounted config files which each generate 6 metrics but all reflect the values from the rootfs, which is already represented. No other xfs file systems will ever be visible to the collectd container so this should be safe. We may want to explore better ways of reporting on the host file system.

csibbitt:
# Testing
* I deployed this config via infrared
* I manually enabled the virt plugin on my compute node (not sure how to do this selectively in the templating)
* Verified that the metrics traffic dropped to reasonable levels
  * compute-0 (w/ 2 VMs): 64 metrics/s (5m avg)
  * controller-0: 55 metrics/s (5m avg)
* Verified sane looking output from each plugin
* EXCEPTIONS:
  * I did not test **`ipmi`** or **`intel_rdt`** as I do not have a BM cloud
  * `ipmi` likely produces an additional ~120 metrics/interval (guessing from source code)
  * `intel_rdt` likely produces ~16 metrics/interval (guessing from source code)


# What about Prom?
* Due to https://github.com/redhat-service-assurance/smart-gateway/issues/42 this is actually horrible
* We'd want to set Prom scrape interval to 5s, but it will still cause staleness of all the slower plugins
* Realistically, setting individual scrape intervals is not currently a great idea
  * We need to fix this bug
  * Or we need multiple scrape targets with different intervals that match the plugins
  * Using a 5s interval for every plugin is calculated at ~123 metrics/s total (not including ipmi, intel_rdt, OR virt)
    * With virt enabled, 10s intervals might be safer

# Next Steps
* I think we should propose that these EITHER these settings, or the same config with global 5s/10s intervals, becomes the upstream default here:
  https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/metrics/collectd-container-puppet.yaml#L80
  and here
  https://github.com/openstack/tripleo-heat-templates/blob/master/environments/metrics/collectd-write-qdr.yaml
  * Upstream actually has CollectdDefaultPollingInterval: 120, obviously hugely conservative compared to this
  * Upstream also has fewer plugins enabled by default, though currently includes tcpconns which more than balances for our additions
  * Not sure if I should I actually submit a change to all these intervals? Especially since they don't play nice with SAF?
  * Conversation with @leifmadsen  is that we both agree we should merge the best version of this (with veriable intervals) and attempt to fix the SG bug before SAF GA.
* Would like more opinions on what is the correct thing to propose upstream, if anything.

tests/infrared/metrics-collectd-qdr.yaml.template
csibbitt: The only xfs mount points the container sees are bind-mounted config files (lots of them!) that each present 6 metrics that are identical to the rootfs ones represented separately (type: overlay). This plugin doesn't do a good job of reporting the df situation on host file systems other than root.

csibbitt:
[Attempt to limit the number of metrics generated by each node](https://trello.com/c/qraEN8KQ/719-attempt-to-limit-the-number-of-metrics-generated-by-each-node)

csibbitt:
@ryan-mccabe appears to be working on the upstream side of this. Hopefully we can figure out the right settings to use upstream. https://review.opendev.org/#/c/694623/2

tests/infrared/metrics-collectd-qdr.yaml.template
```diff
@@ -4,28 +4,47 @@ tripleo_heat_templates:
...
             collectd::plugin::load::reportrelative: true
-            collectd::plugin::interface::ignoreselected: true
+            collectd::plugin::memory::interval: 5
+            collectd::plugin::processes::interval: 5
+            collectd::plugin::uptime::interval: 60
+            collectd::plugin::virt::interval: 5
             collectd::plugin::virt::connection: "qemu:///system"
-            collectd::plugin::virt::hostname_format: "hostname uuid"
             collectd::plugin::virt::extra_stats: "cpu_util disk disk_err pcpu job_stats_background perf vcpupin"
+            collectd::plugin::virt::hostname_format: "hostname uuid"
```
elfiesmelfie: If you want these only in the Compute nodes, you can add a ComputeExtraConfig section.

ExtraConfig gets applied to all roles, but you can use <Rolename>ExtraConfig to apply only to the given role.

tests/infrared/metrics-collectd-qdr.yaml.template
```diff
@@ -4,28 +4,47 @@ tripleo_heat_templates:
...
             - cpu
             - df
+            - disk
             - hugepages
+            - intel_rdt
+            - interface
+            - ipmi
+            - load
+            - memory
             - ovs_events
```
mrunge: ovs should only get installed, when there is ovs running on the node. I'd leave it out for now.
ovs is not the default network in osp anymore since osp15.

csibbitt: Oh! That's really important to know, thanks!

tests/infrared/metrics-collectd-qdr.yaml.template
csibbitt: Discussion on another channel has led to the following:
- The inclusion of `collectd::plugin::virt::*` in ExtraConfig appears to enable the virt plugin even without it being in the plugins list, without breaking the deploy. There are errors in the log but it seems to work fine. This needs to be re-tested from scratch to be sure.
- We think that putting `tripleo.collectd.plugins.collectd: virt` in a `ComputeExtraConfig` stanza is how we'd get it enabled only on Compute nodes. Not tested.

tests/infrared/metrics-collectd-qdr.yaml.template
```diff
@@ -4,28 +4,45 @@ tripleo_heat_templates:
 
 custom_templates:
     parameter_defaults:
-        CollectdAmqpInterval: 1
-        CollectdDefaultPollingInterval: 1
-        CollectdExtraPlugins:
+        CollectdAmqpInterval: 5
+        CollectdDefaultPollingInterval: 5
+        CollectdDefaultPlugins:
```
mrunge: Could you please change this back to CollectdExtraPlugins?

CollectdDefaultplugins will just replace the defaults, ExtraPlugins will append the list of defaults.
Not saying this will cause an issue here, but it lead to an escalation before, as healthchecks rely on the unixsock plugin, which was removed in DefaultPlugins.

tests/infrared/metrics-collectd-qdr.yaml.template
mrunge: I think the latter is the way to go.

leifmadsen: Does TripleO run the `ExtraConfig` first (apply to all nodes) and then the `<role>ExtraConfig` run after? If so then we could do both so that we set the preferred default, then provide an override for `<role>` configs.

tests/infrared/metrics-collectd-qdr.yaml.template
csibbitt: No, I disagree with your thinking here.

This file in specific is the driver for our integration tests and should produce a specific known set of plugins, independent of the upstream list so that tests are on par. No customer is ever going to file an escalation because of this file.

Part of the purpose of this PR is to help @ryan-mccabe decide on exactly which of these should be in the upstream to begin with. That is independent work, but it should be well informed by the discussion here.

leifmadsen: And now this PR makes way more sense to me. OK so the changes here are basically going to be a drop in setup for the upstream file when specifically enabling the "SAF environment" (the file is `environment/enable-saf.yml`).

In this case yes I agree with @csibbitt in that we want to override the generic defaults for collectd because in our case, we want to clearly define exactly what we're enabling and how. 

What would cause an escalation is that someone changes the default plugins on us (which we would otherwise rely on) and enables a plugin that overwhelms the SmartGateway. In that case we'd get an escalation on cloud update if that were to happen.

The environment is being setup specifically for data collection with SAF, so yea, let's make it implicit here so we don't have to debug multiple layers.

leifmadsen: I retract my prior thumbs up and have moved it to another location.

csibbitt: I still don't think we're all on the same page here. I am not trying to defend the use of `CollectdDefaultplugins` in enable-saf.yaml - I think that's up for debate and Leif and Matthias seem to be on opposite side of that debate, currently. I'm undecided. This is the wrong code review for that debate.

I'm defending it's use HERE, in this file, because this is the definition for a test harness and I was specifically disabling some of the upstream plugins in order to achieve the configuration that I want to propose. THIS is the set of plugins I tested, observed, reported on, and am proposing. Not this set plus the default set.

Now there's also a reasonable argument that this file should eventually go away and we should use enable-saf.yaml for our testing (deterministic or not), because that's what the customer will get. That seems right to me, but that file doesn't exist yet, and the change to add it was not even open when I wrote this.

I hope that clarifies what we're talking about.

leifmadsen: Yes that helps clarify it further. In that case, then I think this is perfectly fine to merge and you should do so if there are no further changes to make.

Around the debate as to whether this should match upstream (`enable-saf.yml`) we should probably just setup a small meeting to determine our approach there. I suspect the defaults listed here are going to directly affect Ryan's upstream PR. We should make sure the knowledge from here flows up into that, regardless as to whether we use the `Default` or `Extra` approach. Agreed we can figure that out separately.

I'm going to approve this since I have no further feedback.

mrunge: Okay, following that discussion, use of `CollectdDefaultPlugins` makes sense.

However, I wanted to highlight the general bad practice of using CollectdDefaultPlugins in tripleo yaml files for configuration, unless you exactly know what you are doing, especially when using the product. I'm speaking from experience, I already debugged this during a customer escalation.

PR #101, opened 2019-11-25
Fix perftest for ocp4

csibbitt:
* Minor documentation updates
* Moved dashboard installation to grafana deployment script
* Adjusted naming of everything to match new reality
* Adjusted names and included new pods in dashboard
* Abstracted $all_saf_pods label selector into a dashboard variable
  * Can't make alarms based on these queries though :-(
* Added max/avg and table views for graphs that could benefit
* No testing of events stuff, but from a cursory look it might just work as-is


csibbitt:
@elfiesmelfie I've included you to review here because I want to make sure you can get this grafana instance and dashboard deployed, and that you're comfortable with how to run the performance tests in order to generate traffic. It's the best way I know to see what's going on in SAF, and should be very helpful when we do our HA experiments. 

My thinking is that we'll probably come up with some other dashboard to experiment with, using queries targeting specifically our questions about how things are behaving with 2 or 3 of each component; but this tooling will serve as the entrypoint for that work.

csibbitt:
## Testing

I deployed this and can see all the data I want to see. I can run performance tests, and the results seem nominally similar to what I'd expect from the older OCP3 version of SAF.

## Issues

- There is an annoyance with the dashboard still that I haven't figured out. The queries that count the total number of metrics visible in prometheus sometimes return the error "vector cannot contain metrics with the same labelset". I haven't figured out which part of the query generates duplicate labels because the data is too big when it happens so I can't observe it directly. If I tighten the range in Grafana to JUST capture the test interval and cut off as much "front porch" and "back porch" from the graph as possible, the gauges spring to life and are fine, but if I include too much quiet time (note this is not >0 quiet time, just 'too much'), then it breaks.

- I've also noted that as we move to HA, the stats that require prometheus to scrape QDRs, SmartGateways, and Prom itself are all going to need to move over to PodMonitors instead of ServiceMonitors (do we have that version of the prometheus operator yet?) so that we can see metrics from each component Pod instead of randomly sampling between them by scraping the Service. Not a task for this card PR though, I'll address that as we work our way through the HA work.

csibbitt:
[Migrate telemetry-framework perftest to work with SAF2](https://trello.com/c/hcUaltNb/813-migrate-telemetry-framework-perftest-to-work-with-saf2)

tests/performance-test/deploy/config/minimal-collectd.conf
```diff
@@ -10,7 +10,7 @@ LoadPlugin cpu
 LoadPlugin amqp1
 <Plugin "amqp1">
   <Transport "name">
-    Host "qdr-test.sa-telemetry.svc.cluster.local"
+    Host "qdr-test"
```
pleimer: This works?

tests/performance-test/grafana/grafana-launcher.sh
```diff
@@ -19,11 +19,37 @@ OCP_PASS=$(oc get secret -n openshift-monitoring grafana-datasources -o jsonpath
...
+pod=$(oc get pod -l app=grafana -o jsonpath='{.items[0].metadata.name}')
+echo -n "Waiting for grafana to be listening"
+while ! oc logs $pod | grep -o "HTTP Server Listen"; do echo ...; done
+
+if ! GRAF_HOST=$(oc get routes --field-selector metadata.name=grafana -o jsonpath="{.items[0].spec.host}") 2> /dev/null; then 
+    echo "Error: cannot find Grafana instance in cluster." 1>&2
+    exit 1
+fi
+
+printf "\n*** Creating new dashboards in Grafana ***\n"
```
pleimer: This works, although a more elegant solution would probably be to utilize Grafana's [dashboard provisioning directory](https://grafana.com/docs/administration/provisioning/#dashboards) (the same way data sources are done above).

tests/performance-test/performance-test.sh
```diff
@@ -277,25 +266,25 @@ while true; do
         ;;
         "RESULTS")
             printf "\n*** Collecting test results ***\n"
-            PODNAME=$(oc get pod -l job-name=saf-perftest-notify -o jsonpath='{.items[0].metadata.name}')
-            oc exec "$PODNAME" -- pkill collectd > /dev/null
-            sleep 3
-            oc cp "sa-telemetry/${PODNAME}:/tmp/events.json" /tmp/events.json
+            # PODNAME=$(oc get pod -l job-name=saf-perftest-notify -o jsonpath='{.items[0].metadata.name}')
```
pleimer: Are you disabling events for now because SAF2 doesn't run ElasticSearch yet?

tests/performance-test/performance-test.sh
```diff
@@ -235,20 +224,20 @@ while true; do
...
-                oc logs $(oc get pod -l application=qdr-test -o jsonpath='{.items[0].metadata.name}')
+                #oc logs $(oc get pod -l application=qdr-test -o jsonpath='{.items[0].metadata.name}')
                 STAGE="TARGET"
             fi
         ;;
         "TARGET")
-            TARGETS=$(curl --silent "http://$(oc get route prometheus -o jsonpath='{.spec.host}')/api/v1/targets")
-            QDRWHITE=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"qdr-white"' || true)
-            QDRTEST=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"qdr-test"' || true)
-            PROM=$(echo "$TARGETS" | grep -o '"__meta_kubernetes_service_name":"prometheus-operated"' || true)
```
pleimer: You probably should not delete `|| true` here. If the service monitors do not exist yet, grep will exit with a non-zero code causing the whole test to exit. Given the script is supposed to wait until the service monitor is available, that isn't ideal functionality. Unless you tested this on a blank slate and it works as intended, I am fairly certain this is not what you want

tests/performance-test/grafana/grafana-launcher.sh
csibbitt: Ah, thanks, I was looking for that but ended up just moving the existing code instead. I'll give that a shot!

tests/performance-test/performance-test.sh
csibbitt: I took these out while futzing with something else and never really considered why exactly they were there. I'll put them back, you're totally right, makes perfect sense.

tests/performance-test/performance-test.sh
csibbitt: Yes. As soon as it's working I'll re-test it all and uncomment. I do feel a bit guilty leaving commented code in the file, but it should be very temporary.

leifmadsen:
>     * I've also noted that as we move to HA, the stats that require prometheus to scrape QDRs, SmartGateways, and Prom itself are all going to need to move over to PodMonitors instead of ServiceMonitors (do we have that version of the prometheus operator yet?) so that we can see metrics from each component Pod instead of randomly sampling between them by scraping the Service. 

Yes. We are running Prometheus 0.32.0 now.

> Not a task for this card PR though, I'll address that as we work our way through the HA work. [Edit: On closer inspection this might be wrong, I see metrics have a pod label (though we deliberately drop it from SG scrapes!), so prometheus might be scraping each backend independently]

You mean this part right? (`ServiceMonitor` template)

```
      spec:
        endpoints:
          - interval: 1s
            metricRelabelings:
              - action: labeldrop
                regex: pod
                sourcelabels: []
              - action: labeldrop
                regex: namespace
                sourcelabels: []
              - action: labeldrop
                regex: instance
                sourcelabels: []
              - action: labeldrop
                regex: job
                sourcelabels: []
            port: prom-http
        selector:
          matchLabels:
            app: smart-gateway
```

The metric relabelings were created by someone else when we were doing the dual-SAF setup and I have just forwarded them along. Certainly an area that should be reviewed as I've not looked at what labels we apply recently.

Might be justified to do a label walk of the deployment again for another round of clean up prior to GA release.

csibbitt:
> You mean this part right? (`ServiceMonitor` template)

Yup, that's the part! My hunch is that the pod label is stripped so that metrics scraped from multiple SGs would look "the same" to Prometheus. This will definitely factor into our experiments with multiple SGs.

csibbitt:
@pleimer Fixed!

leifmadsen:
> > You mean this part right? (`ServiceMonitor` template)
> 
> Yup, that's the part! My hunch is that the pod label is stripped so that metrics scraped from multiple SGs would look "the same" to Prometheus. This will definitely factor into our experiments with multiple SGs.

Yea makes perfect sense then. That's basically exactly what we were trying to do with multicast metrics on the bus to 2 SG and 2 Prometheus (keeping the data close to the same on both for replicated data on two databases). Thanks!

PR #102, opened 2019-11-27
Use OCP4 OLM instead of ansible for deployment

csibbitt:
* Removes all the outdated OCP3 deployment machinery
* Proposes a (very POC) replacement quickstart.sh that uses OLM manifests


csibbitt:
[Migrate from `deploy.sh` to OLM + manifests](https://trello.com/c/60X5q4fk/559-migrate-from-deploysh-to-olm-manifests)

deploy/quickstart.sh
```diff
@@ -1,72 +1,71 @@
...
+  startingCSV: service-assurance-operator.v0.1.0
+EOF
+while ! oc get csv | grep service-assurance-operator | grep Succeeded; do echo waiting; done
+oc create -f - <<EOF
+apiVersion: infra.watch/v1alpha1
+kind: ServiceAssurance
+metadata:
+  name: saf-default
+  namespace: ${SAF_PROJECT}
+spec:
```
leifmadsen: I recently changed this `spec` to account for how Ansible does not merge dictionaries very well, so we'll end up with all sorts of issues in the future if we build out more complex dictionaries with defaults, and we don't want to have to define the entire dictionary to override a single value.

So I've flattened the pattern. You'll want to migrate this to:

```
metricsEnabled: true
eventsEnabled: true
```

csibbitt: Ah yes, I saw that, thanks.

PR #105, opened 2019-11-28
Adjust smoketests for SAF2

mrunge:
This is the initial proposal for getting the smoketest back on track for SAF2

Events part not working yet


tests/smoketest/events-smartgateway.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: smartgateway.infra.watch/v2alpha1
```
csibbitt: If we deploy from quickstart.sh, then there will already be smartgateways deployed (called "saf-default-notification-smartgateway" and "saf-default-telemetry-smartgateway". It might make more sense if the smoke test used those ones instead of creating new ones.

tests/smoketest/minimal-collectd.conf.template
```diff
@@ -11,7 +11,7 @@ LoadPlugin cpu
 LoadPlugin amqp1
 <Plugin "amqp1">
   <Transport "name">
-    Host "qdr-white.sa-telemetry.svc.cluster.local"
+    Host "saf-default-interconnect.sa-telemetry.svc.cluster.local"
```
csibbitt: :+1: Just noting that you can shorten this to "saf-default-interconnect"; since containers all have the rest of the name in their DNS search patch (injected by openshift). We don't currently check the hostname on the cert anyways, but currently it shows `subject=/O=cert-manager/CN=saf-default-interconnect` so shouldn't be a problem either way.

tests/smoketest/events-smartgateway.yaml
mrunge: If we are testing already deployed SGs, we don't need this here at all.

tests/smoketest/minimal-collectd.conf.template
mrunge: Neat, you've mentioned this to Paul, but I forgot at the time.

tests/smoketest/events-smartgateway.yaml
csibbitt: That's right, yeah. Ultimately the very most useful thing would be if the smoketest would look at your ServiceAssurance objects and then test what it finds to help customers verify that the topology they've deployed is working correctly. That seems like a big leap from here, though, so just testing the default topology that comes out of the box seems like a good first step for this activity.

leifmadsen: By testing the default scenario we can add other architecture deployment types by just passing a verbose ServiceAssurance manifest with all `*_manifest` parameters.

If a customer has a topology they want or we identify other ones we can pass and save for functional testing.

leifmadsen:
Hey sorry I accidentally resolved a comment from previously and apparently this interface doesn’t allow for unhiding comments easily. The comment was about the DNS settings for the script. The other advantage of using the short name is that now you don’t need to worry so much about the name space. Whatever name space you’re operating in as long as there are objects with that DNS name all your testing should resolve.

mrunge:
> Hey sorry I accidentally resolved a comment from previously and apparently this interface doesn’t allow for unhiding comments easily. The comment was about the DNS settings for the script. The other advantage of using the short name is that now you don’t need to worry so much about the name space. Whatever name space you’re operating in as long as there are objects with that DNS name all your testing should resolve.

No worries, and again: yes, using short DNS names is what we want.
I am undecided yet, if the smoke test should verify other peoples deployments, of if it should be a smoke test for gating functionality.

leifmadsen:
Yea agreed I don't think smoke test is appropriate place for that kind of testing.

I do agree that having smoke test use the saf-default setup makes sense. It also simplifies the smoke test script and gives us a test for the alm-example in the CSV.

csibbitt:
" if the smoke test should verify other peoples deployments, o[r] if it should be a smoke test for gating functionality"

Sorry, I didn't mean to cause confusion here. The primary purpose of this smoketest is to gate changes for CI. My suggestion is that it would be a better test if it used the default SGs that SAF builds rather than building it's own SGs. Anything beyond that is for another time.

mrunge:
On my infra, the tests succeed (minus ES).
Inserting an exit 0 in smoketest_entrypoint.sh just before events_result=0 (line 84), then we'll get a pass.

elfiesmelfie:
`*** [SUCCESS] Smoke test job completed successfully`

Works for me!

leifmadsen:
Are the tests in Travis-CI intended to pass? If that's not possible, then I suggest we disable for the `master` branch until we determine how to test for OCP4 in the upstream CI.

tests/smoketest/events-smartgateway.yaml
leifmadsen: As a follow up and to reaffirm what Chris said, that is work for another time, but by using the built in SGs that the SAO creates on an SA instantiation, then we'll be in better shape for that to eventually become a possibility.

leifmadsen: So this means we want to delete the `metrics-smartgateway.yaml` and the `events-smartgateway.yaml` then correct? We should only pass in a new file that creates a `ServiceAssurance` instance with metrics and events enabled.

mrunge: yes, that's the next step/cleanup

mrunge:
In travis, SAO did not come up after 50 mins. Question is, what prevented if from being deployed.

tests/smoketest/events-smartgateway.yaml
csibbitt: > We should only pass in a new file that creates a ServiceAssurance instance with metrics and events enabled.

quickstart.sh already does this, I don't think it needs to be reproduced here.

Definitely waiting to see events-smartgateway.yaml, metrics-smartgateway.yaml, and the two associated `oc create -f` commands associated with them gone before I'll approve. The results query have been adjusted to look for metrics from "saf-default-telemetry-smartgateway", so these other smartgateway definitions are just noise.

leifmadsen: @csibbitt yea that makes sense, I just meant that the SGs should be created via that method (wherever it may live).

csibbitt:
I think the basic problem is that minishift doesn't have OLM? In any case until there is an equivalent for OCP4, I agree that we should probably just remove the .travis.yml  for now :-(

tests/smoketest/smoketest.sh
```diff
@@ -82,21 +53,19 @@ done
...
 echo "*** [INFO] Logs from smart gateways..."
-for NAME in "${CLOUDNAMES[@]}"; do
-    oc logs "$(oc get pod -l "deploymentconfig=${NAME}-smartgateway" -o jsonpath='{.items[0].metadata.name}')"
-done
+oc logs "$(oc get pod -l "deploymentconfig=saf-default-telemetry-smartgateway" -o jsonpath='{.items[0].metadata.name}')"
 echo
 
 echo "*** [INFO] Logs from smart gateway operator..."
-oc logs "$(oc get pod -l app=smart-gateway-operator -o jsonpath='{.items[0].metadata.name}')"
+oc logs "$(oc get pod -l app=smart-gateway -o jsonpath='{.items[0].metadata.name}')" -c operator
```
leifmadsen: I'm fine with this as it will return as a stream, but the `-c ansible` container will produce prettier output. Just FYI in case a nicer output makes sense here.

tests/smoketest/minimal-collectd.conf.template
```diff
@@ -11,14 +11,14 @@ LoadPlugin cpu
...
```
leifmadsen: Should be able to use the short DNS name here too to avoid hard coded namespace?

mrunge: yes of course. That was an oversight on my side. Sorry.

tests/smoketest/smoketest.sh
mrunge: Alright, with switch to `ansible`

mrunge:
> This worked for me. The events test also appeared to work (I didn't add an `exit 0` line). Here is some events related output:

Thanks for checking.

If you had to add an exit 0, that'd be a bug. It is supposed to work with metrics and events on SAF 2.0. If there `oc cluster up` for OCP4, this should pass on CI unmodified.

PR #107, opened 2019-12-12
Only enable CI testing for OCP3

leifmadsen:
Only enable our CI testing for changes going into the saf-ocp3 branch. Currently
required since we don't have access to an upstream OKD 4 instance that works via
the traditional 'oc cluster up' method. Only instantiation of clusters into AWS
is currently supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,13 @@
 language: minimal
 sudo: required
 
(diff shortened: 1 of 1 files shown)


leifmadsen:
The fact that no tests were triggered proves to me this probably works :)

PR #109, opened 2019-12-20
Initial launch saf on crc script

mrunge:
This is intended to replace minishift-saf.sh at some point.


csibbitt:
Cool, thanks! Only thing I'd add is some notes to _tests/infrared/README.md_ (or even just comment headers in the script itself) explaining what the current limitations are so we don't lose track of what is left to be done with this.

PR #111, opened 2020-01-09
Update quickstart.sh for new application registry

leifmadsen:
We changed the name and path of the application and container registries in order
to avoid named registries. This change updates the quickstart.sh to account for this
change so that a Service Assurance deployment results.

Name of the Subscription has been updated to reflect the pattern that OLM created
for other dependent Subscriptions of the SAO. Updated spec.name to account for the
new package name.

### DIFF
--- a/deploy/quickstart.sh
+++ b/deploy/quickstart.sh
@@ -46,12 +46,12 @@ spec:
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
-  name: service-assurance-operator
+  name: serviceassurance-operator-alpha-redhat-service-assurance-operators-openshift-marketplace
   namespace: ${SAF_PROJECT}
 spec:
   channel: alpha
   installPlanApproval: Automatic
-  name: service-assurance-operator
+  name: serviceassurance-operator
   source: redhat-service-assurance-operators
   sourceNamespace: openshift-marketplace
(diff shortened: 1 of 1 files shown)


deploy/quickstart.sh
```diff
@@ -46,12 +46,12 @@ spec:
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
-  name: service-assurance-operator
+  name: serviceassurance-operator-alpha-redhat-service-assurance-operators-openshift-marketplace
```
csibbitt: Can't we work "Automatic" and "v0.1.0" in there somehow to get the grand slam?

I should have known that teaching people not to use the name for metadata, combined with the natural tendency to want to do that, would lead to this. 

https://i.imgflip.com/3lmr9r.jpg

leifmadsen: I'm only following what OLM already generates for a bunch of the Subscriptions :)

csibbitt: Oh yeah, I'm not blaming you. This is the natural course of things :)

leifmadsen:
[Merge smart-gateway-operator-0.2.0 repo back into smart-gateway-operator and update any build scripts](https://trello.com/c/IDMqkbBp/840-merge-smart-gateway-operator-020-repo-back-into-smart-gateway-operator-and-update-any-build-scripts)

PR #112, opened 2020-01-09
Disable events in main quickstart

csibbitt:
You need to use the ES workaround version for it to work anyways: https://github.com/redhat-service-assurance/telemetry-framework/blob/master/deploy/quickstart_ESworkaround.sh

This also matches the defaults in the SAO: https://github.com/redhat-service-assurance/service-assurance-operator/blob/49e11e5242cfe1debe42da352f12c50659dd4dd9/roles/serviceassurance/defaults/main.yml#L4


elfiesmelfie:
Looks good to me!

PR #114, opened 2020-01-09
Provide way of passing configs to quickstart.sh

leifmadsen:
After I tried to add a third quickstart-style script I decided it was
time to allow ability to pass in configs from a file so that we could
create whatever scenarios we desire for testing.

Sets quickstart.sh to use the existing 'default' configuration
previously. Adds both the ES Workaround and new Quicklab configuration
to a configs/ directory.


deploy/configs/es-workaround.bash
```diff
@@ -70,7 +10,7 @@ spec:
     apiVersion: v1
     metadata:
       name: elasticsearch
-      namespace: "sa-telemetry"
+      namespace: ${SAF_PROJECT}
```
csibbitt: It's my understanding that the certs below only work in `sa-telemetry` namespace. Should add 
`SAF_PROJECT=sa-telemetry` to the top of this file

deploy/configs/quicklab.bash
```diff
@@ -0,0 +1,60 @@
+KIND_SERVICEASSURANCE="apiVersion: infra.watch/v1alpha1
```
csibbitt: Need to hardcode `SAF_PROJECT=sa-telemetry` to match the hardcoded certs

leifmadsen:
I have hardcoded the project for the configurations that provide certificates for ElasticSearch to match what is embedded in the certificates.

Issue #82, opened 2019-10-15
Not deploying cleanly in quicklab


csibbitt:
This is probably more of an issue with the default storageclass config in quicklab, because this works fine on minishift; but I wanted to capture it here so we can track it.

SAF will not deploy cleanly in quicklab because the ES PVC fails to bind.

```
[csibbitt@laptop deploy (master) (⎈ |sa-telemetry/openshift-internal-safperf002-lab-rdu2-cee-redhat-com:443/system:admin)]$ oc get pvc
NAME                                         STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS        AGE
elasticsearch-elasticsearch-cdm-bmr16sp3-1   Pending                                                                                            2m
prometheus-white-db-prometheus-white-0       Bound     pvc-c5f2b95b-ef59-11e9-945a-fa163e0af8aa   2Gi        RWO            glusterfs-storage   2m

[csibbitt@laptop deploy (master) (⎈ |sa-telemetry/openshift-internal-safperf002-lab-rdu2-cee-redhat-com:443/system:admin)]$ oc describe pvc elasticsearch-elasticsearch-cdm-bmr16sp3-1
Name:          elasticsearch-elasticsearch-cdm-bmr16sp3-1
Namespace:     sa-telemetry
StorageClass:
Status:        Pending
Volume:
Labels:        <none>
Annotations:   <none>
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:
Access Modes:
Events:
  Type    Reason         Age               From                         Message
  ----    ------         ----              ----                         -------
  Normal  FailedBinding  1m (x19 over 6m)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
  Normal  FailedBinding  12s (x6 over 1m)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
```

We should be able to work around this by setting an appropriate storage class for elasticsearch_storage_storage_class_name on the eleasticsearch resource. There is no global recommendation for this setting, and your cloud SHOULD take care of setting a default; but we should at least document the quicklab workaround somewhere.

leifmadsen:
@csibbitt try just passing this instead of a full `storageClassName`. This will result in ephemeral storage (probably good enough for testing purposes).

```
apiVersion: logging.openshift.io/v1
kind: Elasticsearch
metadata:
  name: elasticsearch
  namespace: sa-telemetry
spec:
[...]

PR #116, opened 2020-01-16
Adjust quicklab config for latest changes

leifmadsen:
Changes for #82


csibbitt:
Will give it a try!

csibbitt:
:cowboy_hat_face: Merging this untested because the old one is DEFINITELY not going to work, and nothing outside our team dev environments (which are broken) could really be affected by this.

PR #118, opened 2020-01-27
Build nodes list automatically

leifmadsen:
Paul and I were looking at why my dashboard import wasn't returning data even
though the datasource looked right, and connections were working as expected.

Looking further, Paul noticed that the drop down for the nodes wasn't in sync
with the exported_instance values we were seeing in the Prometheus query.
Updating the variable in the dashboard to sync it to what was returned by
Prometheus allowed things to work.

I looked at my old dashboard where I knew I didn't need to do that sync, which
you can originally see imported here: https://github.com/redhat-service-assurance/telemetry-framework/blob/43738e15cf504b31a7a69cc901b2dcdbfe044408/deploy/service-assurance/grafana/configmap-dashboard-definitions.yaml

This change results in the 'nodes' list populating based on the values returned and
not from a static re-generated list.

### DIFF
--- a/deploy/dashboard/rhos-dashboard.yaml.template
+++ b/deploy/dashboard/rhos-dashboard.yaml.template
@@ -2785,11 +2785,7 @@
     "list": [
       {
         "allValue": null,
-        "current": {
-          "selected": false,
-          "text": "controller-0.localdomain",
-          "value": "controller-0.localdomain"
-        },
+        "current": {},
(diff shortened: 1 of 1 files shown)


leifmadsen:
I have not tested this.

leifmadsen:
This isn't needed anymore as I made the same change in the infrawatch/dashboards repo. Closing.

